PR #9, opened 2012-02-07
Http client extension

eyealike:
It would be nice if users of the SDK could customize the HTTP client by subclassing AmazonHttpClient. This pull request has the minimum changes to make this possible. It changes certain methods in AmazonHttpClient from private to protected and it adds a factory method to AmazonWebServiceClient such that concrete clients can customize the AmazonHttpClient instance to be used in the client.

We currently use a custom AmazonHttpClient to use different retry limits based on the exception. For example, we need a high retry limit (say, 20 retries) for "throttling" exceptions, and a low retry limit (like 3) for everything else. Sub-classing AmazonHttpClient lets us do all that.

The amount of customization made possible with this pull-request is still rather limited because most of the state that accumulates during request execution is located on the stack. If there is interest I would be willing to come up with a more substantial refactoring that allows for more extensive customization by encapsulating the request execution state in a separate strategy class.

### DIFF
--- a/src/main/java/com/amazonaws/AmazonWebServiceClient.java
+++ b/src/main/java/com/amazonaws/AmazonWebServiceClient.java
@@ -57,9 +57,27 @@ public abstract class AmazonWebServiceClient {
      */
     public AmazonWebServiceClient(ClientConfiguration clientConfiguration) {
         this.clientConfiguration = clientConfiguration;
-        client = new AmazonHttpClient(clientConfiguration);
+        client = createHttpClient( clientConfiguration );
         requestHandlers = Collections.synchronizedList(new LinkedList<RequestHandler>());
     }
+
+	/**
+	 * Constructs the HTTP client instance to be used in this web service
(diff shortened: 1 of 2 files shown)


fulghum:
Sorry for the lack of response on this one.  

If you're just looking for the ability to control/override retry handling, then check out the setRetryPolicy method in ClientConfiguration:
http://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/ClientConfiguration.html#setRetryPolicy(com.amazonaws.retry.RetryPolicy)

That will give you full control over handling retries.  Let us know if there's more functionality there that you need.

I'm inclined to go ahead and close this PR since there is a supported interface for you to define a retry policy, and also since the repository has changed structure a bit since this was opened. 

I'm interested to hear more about your need for true async IO.  We've passed around the idea a few times, and more customer feedback would be great to hear.  I'll close this PR, but feel free to create a new issue on Async IO where we can discuss those ideas further.

PR #17, opened 2012-03-13
Move test support classes to src/test/java

eyealike:
The 1.3.4 release doesn't build because there are some test support classes (in the new workflow package) that depend on JUnit and Spring Test. Those dependencies aren't declared in the POM. These classes should be under src/test/java, not src/main/java such that we can declare above dependencies with scope 'test' where they don't pollute the main artifact. 

If there is interest, I can additionally have the POM generate an attached JAR with the test support classes.


david-at-aws:
Doing a bit of long-overdue housekeeping. :)

These dependencies have been merged into the flow framework module's pom (as optional, so you won't pick them up by default). I don't think moving the code to the test folder is appropriate; this is not code that's used to test flow itself, it's intended to be used to test workflows developed using the flow framework.

Theoretically it'd be nice to split it into its own module that you can depend on if you want the pick up the test classes, but doing so now would be a breaking change for anyone expecting these classes to come for free with a dependency on `com.amazonaws:aws-java-sdk-swf-libraries`. We'll consider doing that if we ever need to make other breaking changes to flow.

PR #20, opened 2012-05-24
[Improvement] Support type hierarchies while loading objects using DynamoDBMapper

itoshkov:
Suppose we have the following type hierarchy, describing user accounts:

```java
class User { ... }
class Manager extends User { ... }
class Administrator extends User { ... }
```

The data is stored in the same DynamoDB table (e.g. Users). We want to be able to tell the DynamoDBMapper which is the type of each item, so that it is correctly loaded. AFAICT this isn't possible at the moment.

This patch adds the interface TypeDispatcher, which describes a dispatcher (or mapper) between loaded item and Java class. DynamoDBMapper's load, query and scan methods are patched to support such mappers.


david-at-aws:
Closing this as it's drifted pretty far from the current tip of master, but I do like the idea. We'll track it as a feature request, or if someone wants to take a crack at a more up-to-date pull request please feel free.

I'm not sure I would bother with an easy way to use the Java type as the determinator - seems like an unfortunate situation waiting to happen when someone refactors their code. The extra level of indirection also helps make this more reasonable to port to other languages in an interoperable way.

PR #24, opened 2012-09-08
Glacier linear checkums

willglynn:
The existing Glacier upload support does not expose the `x-amz-content-sha256` header. My application uses SHA-256 when transporting data already; ideally, I'd be able to push these precomputed hashes to Glacier and ensure that the stored content matches my expectations, in addition to whatever checksums might be computed locally. Thus, 100ceeccbd0bcaa70c21bcd95b01ca005448018b and 09a0c93a4902784706de3300cee4a153ee0e43b8 add `get`/`set`/`withLinearChecksum` methods to com.amazonaws.services.glacier.model.Upload{Archive,MultipartPart}Request and the corresponding marshalling to `x-amz-content-sha256`.

A separate issue is that UploadArchiveRequest's Javadocs claim over and over again that checksum is a linear checksum. This was a source of confusion -- despite documentation to the contrary, this field is in fact sent as a tree checksum, not a linear checksum. This is addressed in 66c97639b75e0ef01ab4e93cd23b249322505b07.

### DIFF
--- a/src/main/java/com/amazonaws/services/glacier/model/UploadArchiveRequest.java
+++ b/src/main/java/com/amazonaws/services/glacier/model/UploadArchiveRequest.java
@@ -85,6 +85,11 @@ public class UploadArchiveRequest extends AmazonWebServiceRequest {
     private String treeChecksum;
 
     /**
+     * The SHA256 checksum (a tree hash) of the payload.
+     */
+    private String linearChecksum;
+
+    /**
      * The data to upload.
      */
     private java.io.InputStream body;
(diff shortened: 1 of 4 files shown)


fulghum:
Thanks for the feedback @willglynn .  I've verified that the documentation now correctly calls out that the checksum is a tree hash (and not a linear hash):
https://github.com/aws/aws-sdk-java/blob/master/aws-java-sdk-glacier/src/main/java/com/amazonaws/services/glacier/model/UploadArchiveRequest.java

As for exposing the x-amz-content-sha256 header value, this is considered part of the underlying wire protocol, and not part of the service's API (since it's a common requirement for sigv4 signing).  Because of that, having it be populated by the SDK seems reasonable.

If you do want to provide a checksum to ensure data integrity, then you can do so with the tree hash checksum instead of by providing a linear checksum.  Let us know if that doesn't work for you.

PR #48, opened 2013-03-03
Added OSGi manifest header generation to the pom.xml.

paulbakker:
Used the Felix Bundle Maven plugin to generate OSGi manifest headers. This makes the jar file usable as an OSGi bundle. It doesn't impact non-OSGi users at all.

Note that I exported all classes from the JAR. Ideally you would hide implementation classes (by not exporting them) so that nobody accidentally uses them. Because I'm not very familiar with the library yet I can't really decide on this.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
     <modelVersion>4.0.0</modelVersion>
     <groupId>com.amazonaws</groupId>
     <artifactId>aws-java-sdk</artifactId>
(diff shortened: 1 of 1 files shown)


david-at-aws:
Doing some house-cleaning. OSGi is now supported via the [aws-java-sdk-osgi](https://github.com/aws/aws-sdk-java/tree/master/aws-java-sdk-osgi) module.

PR #62, opened 2013-04-14
Use SLF4J instead of JCL (Case 93189881)

saeta:
This is a simple migration to use the simple logging facade instead of using the Jakarta Commons Logging (a deprecated logging framework).

This is in response to AWS Support Case 93189881.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -29,9 +29,9 @@
 
     <dependencies>
         <dependency>
-            <groupId>commons-logging</groupId>
-            <artifactId>commons-logging</artifactId>
-            <version>1.1.1</version>
+            <groupId>org.slf4j</groupId>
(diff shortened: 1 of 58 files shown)


jgangemi:
sad to see this issue is still outstanding.

david-at-aws:
The apache httpclient (which we depend on) is also using commons-logging, so switching the SDK to use slf4j won't get rid of that dependency - it just adds *another* logging dependency. I don't see a compelling reason to switch until/unless they do so we can *actually* jettison commons-logging, but if there is please let us know and we'll consider it for the next time we make a breaking change.

In the meantime, `jcl-over-slf4j` is the way to go if you'd rather have SDK and HttpClient logs piped through slfj4 instead of commons-logging.

PR #76, opened 2013-05-25
Introduce `InstanceFamily` abstraction

nobeh:
Based on the details that are described in this page:
http://aws.amazon.com/ec2/instance-types/instance-details/

InstanceFamilyType introduces a higher level abstraction of the types of the instances that are used. Examples include "general purpose" or "compute optimized". Such abstraction can be better used in application-level awareness of the resources that are required to deliver a type of service. For instance, a service such as data storage needs resources of the family "Storage optimized" which could help hide away lower-level details
of the instance type that is used. It additionally allows to be used in higher-level description of the services that are deployed on the cloud.

A method is added to `InstanceType` to resolve the instance family of the instance type.
There *can* be also a method added to `Instance` that resolves the family directly.

### DIFF
--- a/src/main/java/com/amazonaws/services/ec2/model/InstanceFamily.java
+++ b/src/main/java/com/amazonaws/services/ec2/model/InstanceFamily.java
@@ -1,0 +1,26 @@
+package com.amazonaws.services.ec2.model;
+
+import java.util.Collection;
+
+/**
+ * An abstraction of the family of the instances such as processing bound or I/O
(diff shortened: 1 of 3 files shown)


zhangzhx:
We are auto-generating InstanceType.java based on a shared service model https://github.com/aws/aws-sdk-java/blob/master/aws-java-sdk-ec2/src/main/resources/models/ec2-2015-10-01-model.json#L10082. We would not like to change that file manually. As this PR is almost 3 years, so closing it for now.

PR #139, opened 2013-11-19
Bug fix for marshalling/unmarshalling of set of objects for DynamoDB

4th-line:
When a custom marshaller is used, the underlying value is marshalled
into a string instead of a string set. When the record is unmarshalled
the opposite problem occurs.


david-at-aws:
Sorry for the radio silence on this. I like the feature, but unfortunately it will need to be opt-in to preserve backwards compatibility - either a new annotation or a new parameter to `@DynamoDBMarshalling` that controls whether the custom marshaller is passed the entire set (to be converted into a single string) or individual elements from the set (to be converted into strings which will be stored as a string set).

Closing this pull request since master has changed significantly since it was created. If there's still a need for custom marshaling of Java Set<?> <-> DynamoDB string set, I'd certainly be willing to look at another based on the current code as long as it maintains backwards compatibility for any existing uses of the mapper.

PR #150, opened 2013-12-06
Properties credentials provider from a file location not on the class path.

dipthegeezer:
Hi

I needed to chain various different credentials and noticed that I couldn't do one in which the properties file may sit in 
a file path location not under a classpath. 

So this implements the File credentials as a provider class allowing for it to be added to a chaining of auth providers. 

Thank you


fulghum:
Hey @dipthegeezer , sorry to leave this Pull Request open so long.  This contribution *was* merged into the main codebase a while ago:
https://github.com/aws/aws-sdk-java/blob/master/aws-java-sdk-core/src/main/java/com/amazonaws/auth/PropertiesFileCredentialsProvider.java

Thanks for contributing!

Issue #89, opened 2013-07-17
AmazonS3.createBucket(bucketName) should use region set with setRegion()


axelfontaine:
This makes the interface a lot more natural. Set on region and then expect all subsequent calls to target on that region by default. The current behavior is confusing.

hanshuo-aws:
I believe this has been fixed at the client level:
https://github.com/aws/aws-sdk-java/blob/3baab096839f6848eed40940fb7b6f519a8e0580/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/AmazonS3Client.java#L771-L786

PR #152, opened 2013-12-08
Use getRegion to determine where a bucket should be created when using createBucket(bucketName)

kelcecil:
This addresses issue #89. There is a getRegion method already provided in AmazonS3Client that tries to pull the region from the superclass and defaults back to US_Standard when one is not set. Passing the result from getRegion as the region parameter in CreateBucketRequest is a simple way to implement the desired behavior.


hanshuo-aws:
See #89

PR #155, opened 2013-12-19
Fix support for the EC2 groupSet instance attribute

sjones4:
InstanceAttribute is updated to include security group identifiers in
support of the groupSet instance attribute.

The InstanceAttributeStaxUnmarshaller is updated to unmarshall the
groupSet attribute.


david-at-aws:
Sorry for the radio silence here. This appears to have been added already as of the latest version of the SDK (as `InstanceAttribute#getGroups()`), so closing this pull request.

PR #183, opened 2014-01-31
Add ReadWrite lock to improve performance of getAttributes()

brucej72:
getAttributes() blocks when reading attributes even though this is a
safe operation. This causes blocks in a threaded environment under
load. Added a ReadWrite lock to allow access to the attributes unless
writes are being made.


fulghum:
Hey @brucej72 , sorry we left this pull request sitting around.  The ReadWrite lock changes were incorporated into the main codebase a while back:
https://github.com/aws/aws-sdk-java/blob/master/aws-java-sdk-dynamodb/src/main/java/com/amazonaws/services/dynamodbv2/datamodeling/DynamoDBReflector.java

Thanks for contributing!

PR #189, opened 2014-02-17
Flow Framework: Delegating logic for parsing method arguments to DataConverter

sukrit007:
Description:
Delegating logic for parsing method arguments to DataConverter so that we can control the behavior of de-serialization.

Use Case:
I am using custom Data Converter which does not serialize the java type information in the serialized message.  In order to de-serialize the message, we can derive the type from the method arguments.  However , there is no way to specify in the current implementation.

In order to achieve the same, I  have delegated the call to DataConverter for parsing arguments from the source message.  The behavior for JsonDataConverter is same.

### DIFF
--- a/src/main/java/com/amazonaws/services/simpleworkflow/flow/DataConverter.java
+++ b/src/main/java/com/amazonaws/services/simpleworkflow/flow/DataConverter.java
@@ -14,6 +14,8 @@
  */
 package com.amazonaws.services.simpleworkflow.flow;
 
+import java.lang.reflect.Method;
+
 /**
  * Used by the framework to serialize/deserialize method parameters that need to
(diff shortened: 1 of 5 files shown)


soldiermoth:
:+1:

soldiermoth:
Created this: https://github.com/aws/aws-sdk-java/pull/492 as an alternative if modifying DataConverter isn't desired.

ctrimble:
Are there any updates on the status of this PR?

manikandanrs:
Sorry for delayed response here. The source code is moved to a separate repo owned by the SWF team. Please open a new PR in the [new repo](https://github.com/aws/aws-swf-flow-library).

PR #240, opened 2014-07-04
Added missing services to ServiceAbbreviations: CloudTrail and Kinesis.

Gerrit-Janssen:
Looking at regions.xml these are the only two services that appear to be missing.


manikandanrs:
Thanks for the contribution.
CloudTrail and Kinesis has been added to the ServiceAbbreviations.

PR #242, opened 2014-07-09
Adding the option to modify the end point

guyernest:
Now that Kinesis is available in multiple regions, changing the endpoint is needed. The current code was only applicable for Kinesis streams in us-east-1.


david-at-aws:
Sorry for the (very) slow response on this. I'm definitely on board for making the region configurable here, but the sample in question has been moved around and significantly refactored so I can't directly merge this. I'll put getting region configuration added to the new kinesis sample on our backlog, or feel free to rebase/reopen this pull request.

PR #259, opened 2014-07-31
- swapped commons-logging for slf4j

jgangemi:
the previous pull request is over a year old and the original author hasn't had much activity, so here it is again against the latest version.


david-at-aws:
I'm not opposed, but it doesn't seem to make sense to do this until we make our dependency on Apache HttpClient (which is going to pull in JCL anyways) optional; we'll just be pulling in two logging facades instead of one. In the meantime, I'd recommend using [jcl-over-slf4j](http://www.slf4j.org/legacy.html#jclOverSLF4J) if you want to have both the SDK and HttpClient's logs piped through slf4j.

PR #271, opened 2014-09-12
Migration from json.org json code to Android re-implementation

eethomas:
The Android re-implementation of the json.org code was a clean room reimplementation of the APIs in the json.org code.  Json.org's licence contains the phrase "The Software shall be used for Good, not Evil." which some have interpreted to be a "field-of-use restriction" incompatible with some open source licenses (including the GPL).  This patch removes all code from the json.org codebase and replaces it with the Android-reimplementation (https://apps.fedoraproject.org/packages/android-json-org-java), with a few minor changes to allow compilation with the AWS SDK.

Changes from Android:
- org.json package renamed to com.amazonaws.util.json
- Original JSONStringer.java renamed to InternalJSONStringer.java (api not compatible with json.org version but used internally)
- Created new JSONWriter.java and JSONStringer.java
- Added JSONException constructor with signature JSONException(Exception cause)
- Reimplemented the JSONObject.tryGetString() method (not included in Android version)


manikandanrs:
Hi @eethomas 

Thanks for the contribution. But as @fulghum pointed out, we would want to remove the json.org code completely from the SDK and replace it with Jackson. 

So closing this pull request.

PR #280, opened 2014-10-12
Support newer junit versions for SWF flow tests

bbaetz:
This enables support for newer junit versions.

I've run this against all of the aws examples in the SDK and the recipes from https://aws.amazon.com/code/2535278400103493 using junit 4.7 and 4.11 with no failures.

The issue is that junit 4.7 (which introduced Rules) only wrapped the actual test method. Junit4.8 and above also wrap the Before/After methods. This results in the teardown code being run as part of the async task block, and since there are no dependencies/Promises, that means that it runs before the activity does, leading to incorrect results.

A test for this is the booking test - without this change, the tearDown method is run before the workflow activity runs, resulting in the 'trace' List being null when the activity runs, causing a NullPointerException in the stub activity.

Junit4.11 also changed behaviour so that tests are not run in the same order as in the test file. This exposed issues with tests that use @Test(expected=XXX.class), such as the handleError recipe. The runner is reused, and so the expectedException is kept around from the previous test and passed into the new workFlowTestRule. This is never reset, because possiblyExpectingExceptions only calls setExpectedException when there is an expected exception.

The second scenario happens even with Junit4.7 if you change the order so that testHandleErrorWorkflowWithNonHandleableException is first instead of last in the file.

PS - it'd be nice to have the build-time flow annotation processors available in github/maven too...

### DIFF
--- a/aws-java-sdk-swf-libraries/src/main/java/com/amazonaws/services/simpleworkflow/flow/junit/FlowBlockJUnit4ClassRunner.java
+++ b/aws-java-sdk-swf-libraries/src/main/java/com/amazonaws/services/simpleworkflow/flow/junit/FlowBlockJUnit4ClassRunner.java
@@ -30,6 +30,16 @@ public class FlowBlockJUnit4ClassRunner extends BlockJUnit4ClassRunner {
     }
 
     @Override
+    protected Object createTest() throws Exception {
+        Object testInstance = super.createTest();
+        // Reset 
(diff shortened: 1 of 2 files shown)


manikandanrs:
Sorry for delayed response here. The source code is moved to a separate repo owned by the SWF team. Please open a new PR in the [new repo](https://github.com/aws/aws-swf-flow-library).

PR #283, opened 2014-10-17
DynamoDB tableName override optimization

bsideup:
do not process Table annotation in getTableName method if there is table name override in config


hansonchar:
Hi @bsideup, thanks for the pull request.  Just to clarify, there seems no functional change to this pull request, besides perhaps making it a little faster by avoiding the reflector's call to determine the table name?

On the other hand, it seems to me the existing statement on concatenating the prefix could cause trouble when the prefix is null, since, assuming the tableName is "Foo", concatenating with null would result in “nullFoo”.

hansonchar:
This change will be incorporated in the next release.  Thanks.

PR #326, opened 2014-12-15
Ignore case of S3 custom meta-data headers when getting Object

clicktravel-basharat:
As per the HTTP/1.1 RFC, the message headers should be case-insensitive

http://tools.ietf.org/html/rfc2616#section-4.2

In line with the S3 documentation, the user-defined meta data keys should always be lower-cased:

http://docs.aws.amazon.com/AmazonS3/latest/dev/UsingMetadata.html


mysticfall:
This issue also make it impossible to test user metadata using AWS SDK with https://github.com/jubos/fake-s3 because of this problem : 
* http://stackoverflow.com/questions/2710373/preserving-case-in-http-headers-with-rubys-nethttp

manikandanrs:
Fixed as part of PR https://github.com/aws/aws-sdk-java/pull/706

Issue #328, opened 2014-12-15
KMS: Creating GrantRequest  wit GrantConstrains fails


neoecos:
When creating a grantRequest with a grant subset constraint fails.

The error:

`More than one constraint supplied (Service: AWSKMS; Status Code: 400; Error Code: ValidationException; Request ID: b2b4d06d-84a2-11e4-9a58-5523b2dae570)`

The bug:
When creating a new GrantConstraints() object creates two constraints lists, one for equals constraints and other for subset constraints, even though I'm just using subset constraints. 

request.toString() example:

``` java
{
  KeyId: 531bb400-fd88-44fd-9999-6d2fc2a1000,
  GranteePrincipal: arn:aws:iam::337568200000:user/development,
  Operations: [Encrypt, Decrypt],
  Constraints: {
    EncryptionContextSubset: {contextSubset1=p9nfeie5askl3laeb6j208mt3b0rrl2v, environment=development},
    EncryptionContextEquals: {}  //Added by default by GrantConstraints constuctor.
  },
  GrantTokens: []
}

```

``` groovy
@Transactional(propagation=Propagation.MANDATORY)
    def getEncryptDecryptGrantTokenForContext(def context){
        def request = new CreateGrantRequest().withKeyId(awsKmsKeyId).withGranteePrincipal(userArn)
        request.setConstraints(getGrantConstrainsSubsetForAdditionalContext(context))
        request.setOperations([GrantOperation.Encrypt.toString(), GrantOperation.Decrypt.toString()])
        kmsClient.createGrant(request).getGrantToken()
    }

    @Transactional(propagation=Propagation.MANDATORY)
    def GrantConstraints getGrantConstrainsSubsetForAdditionalContext(def additionalContext){
        def constraints = new GrantConstraints()
        for (Map.Entry context: (awsKmsDefaultContext + additionalContext).entrySet()) {
            constraints.addEncryptionContextSubsetEntry(context.getKey().toString(), context.getValue().toString())
        }
    }
```

PR #329, opened 2014-12-15
Add constraints size validation solves #328

neoecos:
Solves #328 by checking the size of the constraints maps.


fulghum:
Thanks for the feedback @neoecos .  Since this file is generated, we've updated our generators to fix the issue.  You can see the fix in the latest version of the CreateGrantRequestMarshaller.java class:
https://github.com/aws/aws-sdk-java/blob/master/aws-java-sdk-kms/src/main/java/com/amazonaws/services/kms/model/transform/CreateGrantRequestMarshaller.java#L93

Issue #331, opened 2014-12-23
DynamoDBMapper can store List<CustomObject> but not Set<CustomObject>


ccoffey:
Why is it that DynamoDBMapper can marshal a List<User> but not a Set<User> when a Set<User> could easily be stored in Dynamo without loss of precision using a List?

Please see the below example which works perfectly until you replace List<User> with Set<User>, however when you do this you get the following exception.

>com.amazonaws.services.dynamodbv2.datamodeling.marshallers.ObjectSetToStringSetMarshaller: Marshaling a set of non-String objects to a DynamoDB StringSet. You won't be able to read these objects back out of DynamoDB unless you REALLY know what you're doing: it's probably a bug. If you DO know what you're doing feelfree to ignore this warning, but consider using a custom marshaler for this instead.

    @RunWith(SpringJUnit4ClassRunner.class)
    @ContextConfiguration("/application-context.xml")
    public class ProductIntegrationTest {
        @Autowired
        AmazonDynamoDBClient dynamoDBClient;

        @Test
        public void shouldStoreNewProduct() {

            DynamoDBMapper productMapper = new DynamoDBMapper(dynamoDBClient);

            User user1 = new User();
            user1.setUserId(1);
            user1.setFirstName("user1");

            User user2 = new User();
            user2.setUserId(2);
            user2.setFirstName("user2");

            List<User> users = new ArrayList<>();
            users.add(user1);
            users.add(user2);

            Product p = new Product();
            p.setProductName("productA");
            p.setUsers(users);

            productMapper.save(p);
        }
    }

Product

    @DynamoDBTable(tableName = "products")
    public class Product {
        private String productName;
        private List<User> users;

        public Product() {

        }

        @DynamoDBHashKey(attributeName = "productName")
        public String getProductName() {
            return productName;
        }

        public void setProductName(String productName) {
            this.productName = productName;
        }

        @DynamoDBAttribute(attributeName = "users")
        public List<User> getUsers() {
            return users;
        }

        public void setUsers(List<User> users) {
            this.users = users;
        }

    }

User

    @DynamoDBDocument
    public class User {
        private String firstName;
        private long userId;

        public User(){

        }

        @DynamoDBAttribute(attributeName = "firstName")
        public String getFirstName() {
            return firstName;
        }

        public void setFirstName(String firstName) {
            this.firstName = firstName;
        }

        @DynamoDBAttribute(attributeName = "userId")
        public long getUserId() {
            return userId;
        }

[...]

PR #334, opened 2015-01-02
Adding support for marshaling/unmarshaling sets of arbitrary values.

david-at-aws:
See https://github.com/aws/aws-sdk-java/issues/331

Sets of types that are not already stored in one of DynamoDB's native set types are recursively marshaled and stored in a list when ConversionSchemas.V2 is selected. V1 and V1_COMPATIBLE continue to store sets of arbitrary types as DynamoDB string sets (with no way to unmarshal the original objects back out). All schemas can unmarshal data from lists to sets.

This (as a side effect) adds back support for Set<Boolean>, which V1 stored as a number set (1/0 for true/false). V2 dropped support since Set<Boolean> is a little silly, but there doesn't seem to be a compelling reason for going out of the way to block it.

This will prevent us from ever adding explicit mappings from Set<T> to one of DynamoDB's native set types without bumping to ConversionSchema.V3. That seems reasonable to me - the only thing that immediately springs to mind that I'd want to do is maybe Java 8's Temporal types similarly to how we do Date and Calendar? Not sure that there's a good way to add support for those until we drop support for Java 6 and 7 at some point in the far distant future though.


david-at-aws:
Added in [1.9.15](https://github.com/aws/aws-sdk-java/commit/3aca1b9aaa0bb4e783060be699326b0e6a0e4792)

PR #344, opened 2015-01-14
Bug fix for updating dynamodb table from Table object

omerfarukdemir:
Table.updateTable method creates new UpdateTableSpec object without table name parameter for request. So it throws AmazonServiceException, its message like "The parameter 'TableName' is required but was not present in the request".


hansonchar:
Hi @omerfarukdemir, thanks for reporting this bug, which we have actually recently discovered and fixed.  Please stay tuned for the next release.

manikandanrs:
This fix for this is available from [1.9.15](https://aws.amazon.com/releasenotes/Java/7196148869312092)

PR #351, opened 2015-01-26
Modularisation of SDK breaks OSGi bundle imports

brucej72:
The SDK modularisation breaks the OSGi bundle as BNDTools does not have
visibility of all the imports that are required. This results in a
ClassNotFoundException at runtime


brucej72:
I have updated this to fix another problem with the pom.xml. The line:

<Embed-Dependency>*;scope=compile</Embed-Dependency>

needs to have inline=true added to produce a jar file that will work properly in OSGi development in Eclipse

manikandanrs:
Thanks for the contribution.

PR #355, opened 2015-02-05
Add feature to set default profile name via system property for ProfileCredentialsProvider

dai0304:
Each default constructor of API client (ex. AmazonEC2Client) class uses DefaultAWSCredentialsProviderChain.
This chain uses ProfileCredentialsProvider with no argument constructor.

The profile name which this ProfileCredentialsProvider uses is environment variable AWS_PROFILE or "default".
There is no way to control the profile name via system property.

This pull-request enable above.


david-at-aws:
Looks reasonable to me, thanks for the contribution! Can you just confirm that you're submitting this under the Apache 2.0 license?

dai0304:
Sure. I agree.

david-at-aws:
Great, thanks! We'll get this pulled in for our next release.

PR #377, opened 2015-03-09
skipping bridge methods when looking for relevant getters

konrad-krakowiak:
Currently `DynamoDBReflector` isn't ignoring bridge methods which is causing problems when using generic base classes. For example in case of the classes below `User.class.getMethods()` will return 2 versions of the `getId()` method. Correct one returning UUID and bridge one returning Object. Trying to save such user will result in `DynamoDBMappingException` with message _Cannot marshall type class java.lang.Object without a custom marshaler or @DynamoDBDocument annotation_. This change should solve this problem by ignoring bridge methods.

```java
public abstract class Identifiable<ID> {
    protected ID id;
    public ID getId() {  id; }
    public void setId(ID id) {  this.id = id;}
}

@DynamoDBTable(tableName = "user")
public class User extends Identifiable<UUID> {
    @DynamoDBHashKey(attributeName = "id")
    @DynamoDBMarshalling(marshallerClass = UUIDMarshaller.class)
    public UUID getId() {  id;  }

    public void setId(UUID id) {this.id = id; }
}
```

### DIFF
--- a/aws-java-sdk-dynamodb/src/main/java/com/amazonaws/services/dynamodbv2/datamodeling/DynamoDBReflector.java
+++ b/aws-java-sdk-dynamodb/src/main/java/com/amazonaws/services/dynamodbv2/datamodeling/DynamoDBReflector.java
@@ -14,6 +14,8 @@
  */
 package com.amazonaws.services.dynamodbv2.datamodeling;
 
+import org.apache.http.annotation.GuardedBy;
+
 import java.lang.reflect.Method;
 import java.util.Collection;
 import java.util.HashMap;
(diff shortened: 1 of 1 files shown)


hansonchar:
Hi konrad-krakowiak@,

Thanks for the pull request. This change looks incredibly familiar.  After I checked, I realize this exact same issue has been fixed by one of our developers 21 days ago, but somehow didn't make it to the release pipeline.  

  `&& ! (m.isBridge() || m.isSynthetic())`

Will include the changes in the next release.

konrad-krakowiak:
Great! Thanks!

PR #388, opened 2015-03-24
prevent integer overflow when a large number of retries is allowed

amrasarfeiniel:


aws-java-sdk-core/src/main/java/com/amazonaws/retry/PredefinedRetryPolicies.java
```diff
@@ -199,6 +198,10 @@ public boolean shouldRetry(AmazonWebServiceRequest originalRequest,
         /** Maximum exponential back-off time before retrying a request */
         private static final int MAX_BACKOFF_IN_MILLISECONDS = 20 * 1000;
 
+        /** Maximum number of retries before the max backoff will be hit */
+        private static final int MAX_RETRIES_BEFORE_MAX_BACKOFF =
```
hansonchar: This looks a bit convoluted:

```java
	private static final int MAX_RETRIES_BEFORE_MAX_BACKOFF =  (int)(Math.log(MAX_BACKOFF_IN_MILLISECONDS / SCALE_FACTOR) / Math.log(2));
```
Why not just:

```java
	private static final int MAX_RETRIES_BEFORE_MAX_BACKOFF =  6;
```

?

aws-java-sdk-core/src/main/java/com/amazonaws/retry/PredefinedRetryPolicies.java
```diff
@@ -217,7 +221,7 @@ public final long delayBeforeNextRetry(AmazonWebServiceRequest originalRequest,
                 scaleFactor = SCALE_FACTOR;
             }
 
-            long delay = (1 << retriesAttempted) * scaleFactor;
```
hansonchar: The change of bit-wise shift from 1 to 1L also looks unnecessary.  This increase the runtime overhead, albeit very small, with no advantages.  What am I missing ?

aws-java-sdk-core/src/main/java/com/amazonaws/retry/PredefinedRetryPolicies.java
amrasarfeiniel: because if someone changes the other constants, then 6 will no longer be correct.  This value actually depends on the other values here.  In my experience, people will make changes to these when they decide there are better values and not realize they need to change this.  Then you could be sad.

aws-java-sdk-core/src/main/java/com/amazonaws/retry/PredefinedRetryPolicies.java
amrasarfeiniel: it may not be important, but the delay actually is a long, and part of the overflow problem was that it was integer overflowing, not long overflowing.  Also, the perf of bitshifts and multiplies is totally insignificant relative to s3's perf, so that really shouldn't be a concern here.  If you're worried about optimizing perf, you should be looking elsewhere.

aws-java-sdk-core/src/main/java/com/amazonaws/retry/PredefinedRetryPolicies.java
hansonchar: I'd vote for simplicity and change the constant when the other values are changed, instead of having readability to suffer.

aws-java-sdk-core/src/main/java/com/amazonaws/retry/PredefinedRetryPolicies.java
hansonchar: I'd vote to keep it simple and keep the 1 instead of 1L.  Nothing to do with optimization, but more  to simplicity and readability.

hansonchar: Hi @amrasarfeiniel, may I withdraw my original comment/vote?  Just realized I misread the code.  I know this is minor, but 1L is nicer and more correct, and reflect the intent (of using the long type) more closely.  My apologies.

hansonchar: Hi @amrasarfeiniel, may I withdraw my original comment/vote.  Just realized I misread the code.  I know this is minor, but 1L is nicer and more correct, and reflect the intent (of using the long type) more closely.  My apologies.

aws-java-sdk-core/src/main/java/com/amazonaws/retry/PredefinedRetryPolicies.java
amrasarfeiniel: ok, if I do that I'll also add a comment on the other constants about how people need to update this new one if they update the others and give them the formula I used for it.  Is that better?  I'm definitely not committing the magic number 6 with no context.

hanshuo-aws:
@amrasarfeiniel The change looks good to me. As usual, I would need you to confirm that you contribute this change under Apache 2.0 license. Thank you!

amrasarfeiniel:
yep, apache 2.0 is good by me.  Maximum permission for others to do whatever they want with this :)

PR #395, opened 2015-03-31
split DynamoDBMapper batch write into two public steps

trasa:
We want to use the DBMapper to handle the details of bridging DynamoDB to Java objects, but also want to batch up requests that span DynamoDB Tables for performance reasons. DynamoDBMapper provides all the functionality we need; we just need to expose that functionality.

The reason for this is so that we can call these two parts independently, so that we can use one mapper to write batch requests for multiple tables.

I made this method public; this effectively breaks writeBatch into two parts: create the items, then write them.

```
// get requests for player and playerWallet mapped objects
DynamoDBMapper.BatchWriteRequests playerRequests = mapper.createBatchWriteRequests(playerList, Collections.emptyList(), someConfig);
DynamoDBMapper.BatchWriteRequests walletRequests = mapper.createBatchWriteRequests(playerWalletList, Collections.emptyList(), someConfig);

// add the two sets of batch requests together
playerRequests.add(walletRequests);

// transmit the entire set
List<DynamoDBMapper.FailedBatch> failedBatches = mapper.writeBatchRequests(playerRequests);
```


trasa:
For what it's worth, I've applied this change forward to my 1.10.45 branch: 
https://github.com/trasa/aws-sdk-java/tree/1.10.45.multiBatch

shorea:
Not sure I understand the goal of this PR. The mapper allows calling batchWrite with a bag of objects so you can have different POJOs for different tables.

trasa:
Yeah, this change made more sense for the project it originated in - batchWrite(List<Object>)  is certainly viable.

shorea:
Okay gotcha. Thanks for confirming. Closing the PR.

PR #406, opened 2015-04-14
Sign request after retry pause, to avoid "RequestTimeTooSkewed" errors

alewando:
Current logic in AmazonHttpClient generates a request signature prior to performing any pauses prescribed by the configured RetryPolicy. The request timestamp is included in the signature, so generating it before pausing for retry can result in a significant difference between the request timestamp in the signature and the time AWS actually sees the request, resuling in "RequestTimeTooSkewed" errors. This PR modifies the logic such that pauses are performed before signing the request.


manikandanrs:
Hi @alewando 

Thanks for the contribution. We will get this merged in the next release. Can you confirm if you are submitting it under Apache 2.0 License

alewando:
Yes, Apache 2.0 License is fine.

PR #415, opened 2015-04-23
Add option to size the response metadata cache.

rmuir:
In some long-running applications, this cache holds entries just long enough to create old generation garbage.

As the cache is only used for diagnostic purposes, it would be great to configure the size (e.g. to zero) when it causes problems.

See also https://github.com/elastic/elasticsearch-cloud-aws/issues/193


manikandanrs:
@rmuir 
Thanks for the contribution. We will get this merged as part of our next release.

PR #417, opened 2015-04-28
Removal of JSON.org JSON library code

linead:
I've removed the JSON.org JSON library code and replaced it with calls to Jackson. 

The only class with the same name is the JSONWriter which been re-written to use Jackson as well, it now only shares a public API with the original.

This means there's no need for an exception in the LICENSE.txt, such usage restriction makes the license incompatible with The Open Source Definition and The Free Software Definition

### DIFF
--- a/aws-java-sdk-acm/pom.xml
+++ b/aws-java-sdk-acm/pom.xml
@@ -5,40 +5,31 @@
   <parent>
     <groupId>com.amazonaws</groupId>
     <artifactId>aws-java-sdk-pom</artifactId>
-    <version>1.10.58</version>
+    <version>1.10.59-SNAPSHOT</version>
   </parent>
   <groupId>com.amazonaws</groupId>
   <artifactId>aws-java-sdk-acm</artifactId>
   <name>AWS Java SDK for AWS Certificate Manager</name>
   <description>The AWS Java SDK for AWS Certificate Manager module holds the client classes that are used for communicating with AWS Certificate Manager service.</description>
   <url>https://aws.amazon.com/sdkforjava</url>
-
-  <!-- The dependencies section in pom.xml is auto generated. No manual changes are allowed -->
   <dependencies>
     <dependency>
-        <artifactId>aws-java-sdk-core</artifactId>
-        <groupId>com.amazonaws</groupId>
-        <optional>false</optional>
-        <version>1.10.58</version>
+      <groupId>com.amazonaws</groupId>
+      <artifactId>aws-java-sdk-core</artifactId>
+      <version>1.10.58</version>
+      <optional>false</optional>
     </dependency>
-</dependencies>
-
+  </dependencies>
   <build>
     <plugins>
       <plugin>
-        <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-compiler-plugin</artifactId>
       </plugin>
-
       <plugin>
-        <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-javadoc-plugin</artifactId>
       </plugin>
     </plugins>
   </build>
-
-  <!-- This profile uses the JAPICMP plugin to generate a report of changes between the release version and the latest version -->
-  <!-- For more information on the plugin, see https://github.com/siom79/japicmp -->
   <profiles>
     <profile>
       <id>versiondiff</id>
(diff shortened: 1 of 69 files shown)


adrianbk:
+1 for this.

ebourg:
It would be nice to merge to patch, the non-free license prevents Debian and Fedora from distributing the AWS SDK.

manikandanrs:
@ebourg 

Unfortunately we wouldn't be able to merge this pull request in its current form. Most of the SDK client code is generated. We will need to make changes to generator templates before removing the JSON org code. 

Can you give us more information about the "non-free license prevents Debian and Fedora". The SDK is open sourced under Apache 2.0 license. Is that not enough ?

ebourg:
@manikandanrs 

The SDK includes the JSON processing classes from [json.org](http://json.org) in the `com.amazonaws.util.json` package. These classes have a MIT license with the additional clause _"The Software shall be used for Good, not Evil"_. This is a field of use restriction violating the rule 6 of the [Open Source Definition](https://opensource.org/osd) : _"The license must not restrict anyone from making use of the program in a specific field of endeavor"_. This clause turns the license into a non-free one, and as such Linux distributions like Fedora and Debian are not allowed to distribute it.

This may look like a joke but sadly it isn't, the json.org author Douglas Crockford is very serious about this and repeatedly refused to remove it. That's why most open source projects using the json.org code see pull requests like this one to replace the offending code with one of the many alternatives available.

> Most of the SDK client code is generated. We will need to make changes to generator templates before removing the JSON org code.

Is the code generator open sourced and available somewhere?

shorea:
Looking into this. We actually moved off of the JSON code in 1.10 and we are only referencing in legacy code and tests so it should be easy to remove.

shorea:
I misspoke, we have only migrated off it for exception unmarshallers. Apparently we still rely on this code heavily for request/response marshaller/unmarshallers. This will be a bigger change as it has the potential to affect every JSON based service.

manikandanrs:
@ebourg 

Thanks for the explanation. We will take a look at this and see if we can remove the JSON org code from the SDK.

To answer your question on the code generator. Yes, it is open sourced. It is available as a module in the repository. 
https://github.com/aws/aws-sdk-java/tree/master/aws-java-sdk-code-generator

The code generator exists as a maven module but none of the service modules consume it via maven. We have plans of exposing a maven plugin for the code generator and use it for the service modules.

fulghum:
@ebourg We just open-sourced the generator code as part of this repo:
https://github.com/aws/aws-sdk-java/tree/master/aws-java-sdk-code-generator

We still need to do some more work to make the generator easy to run externally, but you could take a look at the Freemarker templates if you're interested in contributing.  We'd love to drop the Json.org code and clean up this licensing issue.

linead:
@fulghum I'd be happy to make this changes in the generator. Does that mean that the code here is generated and then committed to the repo?

shorea:
We may be able to get away from having to make code generator changes. Talking with legal to see if we can use JSONWriter interface much like this original PR does. If that's the case this task will become much easier. Hang tight until then, should have an answer soon.

shorea:
So unfortunately I don't have better news for you but here is where we are.

We do not feel comfortable reusing JSON.org's JSONWriter interface without distributing their license. Since this code has been in the SDK for years we feel it's an unnecessary risk of copyright infringement. So the plan is to re-implement entirely with Jackson and the marshallers would have to be updated to reflect Jackson's interface. Because we make guarantees about backwards compatibility within major versions we cannot drop the JSON.org classes which means we can't drop the license until the next major revision. 

So for next steps, I can drive transitioning over to Jackson for our marshallers and get it prepped for our next major version. Once we rev the version we can then drop the JSON.org classes and we will be good to go.

linead:
Could we amend this PR to remove that interface would that be OK?

shorea:
We unfortunately cannot do that at this time. We guarantee that older clients will work with the latest version of the core. So say you are using version 1.10.30 of Dynamo DB and 1.10.50 of the core then it should be compatible. Removing the interface would break this contract.

ebourg:
The next major version would be 1.11, right? If an official patch could be prepared on a branch and later merged into the next major version, that would allow the Linux distributions to pick it up and distribute the SDK. Even if it breaks the backward compatibility the Fedora/Debian maintainers can fix the issues in the packages using the SDK.

shorea:
JSON.org has been removed in 1.11.0

ebourg:
Thanks a lot, this looks perfect now.

PR #425, opened 2015-05-04
AmazonSQSAsyncClient documentation fix.

gatorcse:
The no-arg and basic credentials constructors both claimed to be blocking. They both call other constructors that claim to be non-blocking (and this is the async client) so the docs have been changed to reflect this.


breedloj:
Thanks for the submission!  The documentation in question has been corrected.

PR #432, opened 2015-05-13
Add possibility to use SSL with a SecurityManager

rmuir:
verifyMasterSecret() reflects on a sun internal, protected package to look at the master secret.
It catches various reflection-related exceptions, but doesn't handle the case where access is denied
by a security policy RuntimePermission("accessClassInPackage.sun.security.ssl"), and doesn't run its
code in a doPrivileged block so that a user can choose to restrict access to this permission just to
the AWS core SDK jar codebase.

With this change users of the API have choices available: to not grant the access at all, or to grant
the access to just the AWS jar. Users not running running with a SecurityManager enabled are not impacted.


aws-java-sdk-core/src/main/java/com/amazonaws/http/conn/ssl/SdkTLSSocketFactory.java
```diff
@@ -130,26 +133,36 @@ public Socket connectSocket(
...
+                    } catch (PrivilegedActionException e) {
+                        // unwrap checked exception
+                        Throwable cause = e.getCause();
+                        if (cause instanceof InvocationTargetException) {
+                          // unwrap checked exception
+                          cause = cause.getCause();
+                        }
+                        failedToVerifyMasterSecret(cause);
                     }
+                    if (success && masterSecret == null)
```
hansonchar: We can get rid of the "success" variable if we moved line 164-165 to line 152.

rmuir: This method will catch its own securityexception if we do that. This needs to be differentiated between that of an actual securityexception coming from doPrivileged(), which is not a checked exception, hence will not be wrapped in PriviligedActionException.

rmuir: Personally, i wouldn't throw a SecurityException here at all, then you could do the simplification you suggest. Its absolutely the wrong exception for AWS logic to be throwing from this method: its "Thrown by the security manager to indicate a security violation."

hansonchar: I see why the need for the success flag now.

On SecurityException, even though It's not thrown by the security manager, it's a security violation.  I agree this is probably the wrong exception, but not sure if there is a better one.

rmuir: I wanted to not "break the api" and change the exception type, to make this change less invasive, to not have any impact on backwards compatibility, or other concerns. Personally, i think IllegalStateException is a good one for this case though.

SalusaSecondus: This should probably throw an SSLException as it is a clear violation of the protocol.

hansonchar: Thanks @SalusaSecondus for the suggestion (which I personally concur).

In any case, this pull request is currently under security review, and will likely be merged with other more recent changes introduced to this part of the code once we get the green light from the security team.

rmuir: @hansonchar shall I defer the two suggestions proposed here by @SalusaSecondus ? They are good improvements, but I don't want to interfere with a security process. I could also make a second PR following up with those changes if that is easier.

hansonchar: Hi @rmuir, all sounds reasonable. In any case, I'd suggest to wait for the next release, and then decide where to go based on the latest source there.

SalusaSecondus: The next release is out (Java 1.9.40). Let's revisit this change based on the new source.

rmuir:
I updated the pull request with a simpler approach. Basically just to catch the security exception.

Personally i do not think anyone who understands java security would want to grant the AWS jar permission to do what is happening here. So I removed the AccessController.doPriviledged piece: I think that would be better done as a followup, if its really needed and really made me too nervous.

So hopefully this is easier to digest, and it allows me to use AWS libraries without granting this extremely dangerous permission, which is a big security hole.

cc @hansonchar @SalusaSecondus

s1monw:
Is there any way to speed this up here. It basically prevent to properly lock down users of this API with security manager which is a shame and the fix seems pretty straight forward to me. I wonder what keeps this PR from being merged though. We are currently hack around these problems but it would be nice to get this in at some point

rmuir:
You also cannot even do this broken stuff anymore under java 9. So you must fix it!

SalusaSecondus:
@s1monw Let me grab the right people for this. While I *can* merge this myself, I'm not the primary owner of this code so I don't want to directly muck around with this (unless they give me the go-ahead).

rmuir:
I would recommend simply removing this verifyMasterSecret stuff completely, so that aws-sdk-java will work with java 9 modularization (jigsaw). this code is a security hole...

SalusaSecondus:
@rmuir Unfortunately, I'm not familiar yet with Java9's changes. Ignoring the performance hit of the exceptions, what is the security hole with this code? It will run even if denied access (so you can lock it down under a security manager).

s1monw:
> Ignoring the performance hit of the exceptions, what is the security hole with this code? It will run even if denied access (so you can lock it down under a security manager).

if you look it down you will hit a `SecurityException` right here https://github.com/aws/aws-sdk-java/pull/432/files#diff-fc92d6a182070d94f17b49e2e18481a4L186
which causes the code to barf as far as I can tell. how do you thing this can be locked down? Performance of an exception is not an issue here...

SalusaSecondus:
I'm feeling a bit more confused. I'm currently speaking about the proposed patch, not the existing code.

If the permission is granted, then we check for a null master secret which certain odd and buggy TLS clients can be forced to accept by attackers. I'd need to check, but I'm pretty sure that some older versions of Java (which people still use) are vulnerable to this. So this check, if it can be executed, protects people from that issue.

If the permission is not granted, then one of several Exceptions will be thrown and caught.  The caught exception will then be passed to failedToVerifyMasterSecret which emits logs, but doesn't propagate the exception.

Either way, this seems to be acceptable.

s1monw:
> Either way, this seems to be acceptable.

I agree I wonder if these JVM are EOLed years ago? When do you stop supporting the ancient stuff :)

SalusaSecondus:
> I agree I wonder if these JVM are EOLed years ago? When do you stop supporting the ancient stuff :)

While I may agree with you on that, I'm not the person who really makes that call. I'm very aggressive on upgrades and deprecating old systems. That probably has something to do with my role as a security engineer. :smile:

rmuir:
Lets see, you violate the rules of java and access a *restricted package*. I guess you still don't seem to understand that you just cannot do this as of java 9. Try it, download a build from http://jdk9.java.net/jigsaw, and you will see, the whole library breaks with yet another exception that you don't catch.

No users of this library are able to prevent this, if they use this library is forced to allow it *globally* since you don't do it in an AccessController block: this means arbitrary code has permissions to sniff ssl connections and all kinds of bad things because this permission must be granted for the library to work.

The only thing someone can do is an uberhack like I did (sneaking-throwing an exception that you *do* catch from a custom Policy impl) but that is really over the top.

Seriously, you guys need to fix this. I am not afraid to make a blog, make sure everyone knows not to trust their data with you, open a CVE about it, fork a secure version of this library whatever it takes. If you want a shitstorm, I can bring one.

s1monw:
> While I may agree with you on that, I'm not the person who really makes that call. I'm very aggressive on upgrades and deprecating old systems. That probably has something to do with my role as a security engineer. 

I hear you - I really think this is dangerous as it is today. These things are private to the Java library for a reason so you shouldn't access the symmetric key to begin with I wonder what the bigger security risk is here at this point.

SalusaSecondus:
@rmuir I get this and totally get that we need to fix this. We are trying to figure out the *best* fix to this. There is a reason that I'm aggressively wrangling the package owners over here.

s1monw:
@SalusaSecondus ++ apply some pressure :)

SalusaSecondus:
@s1monw @rmuir A quick update:

* I have engagement from the team
* I have found the original commit of this code
* We are (re-)investigating the implications of the bug this is working around
* We are still evaluating the proper solution. (If you have recommendations as to the proper solution, other than simply saying "Don't check this value", I'd love to hear them. The reason that I don't want to hear "Don't check this value" has nothing to do with the correctness of that solution, simply that I am already aware of it and have it on our list to consider. :smile: )
* **We will not drop this.**

rmuir:
> We are still evaluating the proper solution. (If you have recommendations as to the proper solution, other than simply saying "Don't check this value", I'd love to hear them. The reason that I don't want to hear "Don't check this value" has nothing to do with the correctness of that solution, simply that I am already aware of it and have it on our list to consider. :smile: )

It is extremely simple, it is your only option. You will not be able to do this on a java 9 virtual machine. 

There is nothing to "consider".

SalusaSecondus:
(The following links are provided as notes to myself and colleagues)
* [InaccessibleObjectException](http://cr.openjdk.java.net/~mr/jigsaw/spec/api/java/lang/reflect/InaccessibleObjectException.html)
* [JEP 261](http://openjdk.java.net/jeps/261)
* [JSR 376](http://openjdk.java.net/projects/jigsaw/spec/)

SalusaSecondus:
Just a quick update. While I'm not promising daily updates, I'm going to work to keep you aware of where we are.

* I have duplicated the issue in Java 9. Thank you for the leads and links.
* I'm busy swapping code with the team in question
* While the domain expert wasn't available yesterday or today, we expect him to be around tomorrow. Hopefully he still has some of the original documentation and test cases which resulted in this weird hack to begin with.
* My current straw-man is based on @rmuir's PR with additional logic to make everything work in Java 9. It should also be more future proof.
* I don't (yet) have an ETA which I can commit to. Sorry.
* I am watching this thread pretty closely so if you post here, I'll see it and respond to the best of my ability.

shorea:
I thought the domain expert on this issue would be back today but unfortunately he won't be around until next Wednesday. We can proceed without him though as it looks like the fix is known and we can consult him when he returns to see if this workaround can be removed completely.

We propose keeping the existing check to verify the master secret, wrapping it in a doPrivledged block per @rmuir original request to allow security focused customers to grant (or restrict) access to only the Java SDK, and not doing the check at all for Java 9 since it won't work and shouldn't be needed (I assume whatever bug this workaround is for has been since fixed although I can't find any official documentation on the bug so not entirely sure). @SalusaSecondus has been kind enough to provide psuedocode for the proposed approach.
```
boolean isGood = doPriveleged() {
  try {
    if (javaVersion > 1.8) {
      return true;
    } else {
      // Do current and return result
    }
  } catch (Exception ex) {
    return true;
  }
}
```

Let me know if you would prefer to make these modifications yourself (assuming you agree with them) or if you would rather us build on top of your pull request to make the changes.

rmuir:
So you create a complex scenario with conditionals around jdk version combined with doPrivileged, for a phantom bug you dont even know what it is?

way to be secure.

SalusaSecondus:
@rmuir We know the immediate bug and have records of it being reproduced (and this patch preventing reproduction).  Having a null master secret on a TLS connection is a "very bad thing" and should never be possible. Unfortunately, at at least one point in time it was. We're still working to dig up the old test cases and documentation to determine which (if any) modern JVMs are impacted by the bug this is working around.

I've looked over some of the original discussion and I have to agree that checking this value and blocking this TLS misconfiguration was the right thing to do. We shouldn't have had to do this in the first place, but unfortunately Java has bugs which we sometimes need to work around.

rjernst:
> We shouldn't have had to do this in the first place, but unfortunately Java has bugs which we sometimes need to work around.

Here's a simpler approach (in pseudocode):
```
if (using ssl && jvm is known to be buggy) {
  throw new RuntimeException("JVM version x.y.z is known to have issues with ssl. Please upgrade to a newer version.");
}
```

rmuir:
Here is my general frustration. I am trying my hardest to lock down elasticsearch, so the problems in the past (e.g. thousands of zombied machines on AWS) never happens again.

But things like this, are huge holes that make it hard for me to do this. It means i have to leave open giant holes that allow any code to hijack/sniff/whatever SSL connections. That is terrible!

Another example is the messy serialization of InternalConfig$Factory which requires `suppressAccessChecks`, defeating any attempt of ours at protecting confidential data like aws keys. I will open another PR for that one.

I am just saying, it is totally clear this code has never been used in a secure environment, and i think given the shit that has happened in the past, we should all agree that security should be a goal and we should move towards that and not just leave stuff hanging. 

Otherwise these things will repeat themselves.

shorea:
I don't see why the proposed approach prevents you from locking down access to the session's master key. If you don't grant the access we'll just log a debug statement and carry on.

rmuir:
Right but with java 9, you will get a different exception for anyone not running with security manager.

rmuir:
and there is no justification for the current hackery, no code comments for the crazy reflection of sensitive internal classes, etc. I would really just remove this.

shorea:
I agree that it shouldn't be necessary but unfortunately at this time I don't have enough information about which jvms are affected to feel comfortable completely removing it. We have to consider the security implications of the customers we have on older jdks. I think the proposed approach makes the code functional on java 9, gives security focused customers more flexibility and control on java 8 and below, and still keeps the workaround for those that may need it. I don't want to be hasty in removing this hack, and yes I agree it's an ugly hack, for the sake of cleaning up code.

SalusaSecondus:
> Right but with java 9, you will get a different exception for anyone not running with security manager.

@rmuir, this is why we're are proposing simply skipping the check on Java 9. As I mentioned earlier, I am on the extreme side of "Patch everything, run the latest version" but I've also seen enough cases where people (or large companies) can't move everything over. So, instead of just running on a bad JVM, they'll run on a bad JVM and an old version of our SDK because it doesn't support the older versions. Where we should draw the line is a conversation for people other than me. I've provided my input (and I know that it's been listened to), but fundamentally, it is not a simple question.

Now, for "InternalConfig$Factory", can you give me a few more pointers here? If you're having problems with this class (including running it under a security manager), please let us know. While I'm certain we can track it down, having a bit of a description will help some of our other engineers.

shorea:
@rmuir do you feel comfortable proceeding with the proposed approach?

SalusaSecondus:
@rmuir We would really appreciate your feedback on this approach.

rmuir:
I don't think the current unjustified code should be there at all: I stand by my guns on that. 

When it comes to things like peeking at internal sun libraries, the onus is on the one doing the hackery, not the other way around. You have no business messing with this!

If you insist on doing hackery to workaround XYZ version of java, then contain such hacks to those versions or below. The default for *ancient* stuff can always be to simply fail. Such workarounds (especially ones with severe implications, like meddling with internal libraries) can be allowed only when a system property is passed.

But I'm guessing you dont even know what versions of java this hack applies to. If you did, there would at least be a code comment. When dealing with jvm bugs, there should at least be a link to the java bug number, even if it is a security issue and is not public.

Just like we do here: https://github.com/elastic/elasticsearch/blob/master/core/src/main/java/org/elasticsearch/bootstrap/JVMCheck.java

As far as the `InternalConfig$Factory`, just look at the jackson deserialization of this config: because the necessary constructors/classes are not public, they cannot be instantiated by jackson-databind code (which is in a different package). So jackson-databind only works if you allow `RuntimePermission suppressAccessChecks`. We cannot allow this (see this thread if you think its not serious: http://mail.openjdk.java.net/pipermail/hotspot-dev/2015-June/019277.html), at the least, because it destroys the ability to protect confidential data at all in our system. When i look at that code, i see unjustified overengineering of this configuration, which looks like it could be something way simpler, like a .properties file, instead of tons of classes with json serialization that requires security permissions "as dangerous as Unsafe". Maybe it was written by an intern, I do not know, but that hole should also be closed.

SalusaSecondus:
@rmuir Thank you. I do understand your position. However versions of the SDK which we have decided to support do have this bug. (If I had a CVE or something I'd post it. I'm having some trouble digging up the original issue for various, mostly stupid, reasons.) I do, however, understand what this is defending against and continue to believe it is worth the fix it provided that it:
A)  Can be run under a security manager, AND 
B) Not cause problems in Java9, AND
C) (optionally but likely) be able to be turned off via a system property.

I'd also like to improve the comments around it so that people can better understand that why of the code. That might be the hardest part unfortunately.

As for the "InternalConfig$Factory problem, have you created an issue with this which I can reference and track? I (personally) believe that our code should be runnable under a security manager, so if it's not we need to fix that.

shorea:
@rmuir We were able to track down details of the vulnerability.
http://www.oracle.com/technetwork/topics/security/cpujan2015-1972971.html
https://access.redhat.com/security/cve/CVE-2014-6593
 
Since we know now the affected versions we can restrict this workaround to just those versions. This vulnerability has been fixed in the following JDK versions.
Java SE 6 Update 91
Java SE 7 Update 51
Java SE 8 Update 31

Here's the updated pseudocode.
```
boolean isGood = doPriveleged() {
  try {
    if (!isJvmAffected()) {
      return true;
    } else {
      // Do current and return result
    }
  } catch (Exception ex) {
    return true;
  }
}
```

If you are satisfied with this approach I can begin implementing later this week.

rmuir:
@shorea yes this would be improved, but just remember "do current" as-is is still broken, it fails to catch SecurityException when trying to apply the hack.

That is the whole essence of the current problem under security manager.

shorea:
The blanket try/catch should take care of the security exception. The "do current" will be changed to be best effort. If we can't verify the master secret for any reason, including use of a security manager, we'll just return true and proceed as normal.

rmuir:
@SalusaSecondus I opened this issue about `suppressAccessChecks`: https://github.com/aws/aws-sdk-java/issues/528

SalusaSecondus:
@rmuir Thank you. I'm watching that issue as well now. I definitely want us to be able to use the SDK in the most secure ways (and situations) possible.

SalusaSecondus:
As it's been two weeks, this is just a quick update letting you know that we haven't forgotten this. We're taking the time to do this right and give us a better path forward to finding (and fixing) any other SecurityManager problems which may exist.

shorea:
This particular issue has been fixed in the latest SDK. Still looking into Issue #528 and we are still exploring how we can improve our testing and documentation around using the SDK in a security manager.

rmuir:
@shorea thank you for fixing this! unfortunately, we are met with a bigger problem in recent releases, I opened an issue for that: https://github.com/aws/aws-sdk-java/issues/552

Issue #436, opened 2015-05-28
headers parsing should be case insensitive


jackphel:
I'm using https://github.com/jubos/fake-s3 as an offline server for development and testing, but because https://github.com/aws/aws-sdk-java/blob/master/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/model/ObjectMetadata.java#L625 is only looking for the capitalized version of the header, it misses the "etag" header that this dev server always returns as all lowercase.

hansonchar:
One work around is to make use of:

```java
Map<String, Object> map = ObjectMetadata#getRawMetadata()
```
or
```java
Object value = ObjectMetadata#getRawMetadataValue(String)
```
to access the underlying raw headers with lower case names if necessary.

Even though these methods are intended for internal use, it seems they could work well against the fake server in this peculiar case.

hansonchar:
Will test on the pull request.

hansonchar:
Turns out this patch causes many of our existing integration tests to fail, and evidently not safe. For now, I'd suggest to make use of the work around as suggested earlier.

jackphel:
Can this stay open, though? The header parsing still assumes that the headers have a particular case, even though http headers are case insensitive. So, running the java client against a mocked s3 service might fail unexpectedly depending on the case.

"_Field names are case-insensitive_" - http://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html

hansonchar:
I'd prefer to re-open this issue when there is a safer patch.  For now the cost doesn't seem to justify the benefit.

jackphel:
I think I'm just going to put a proxy around https://github.com/jubos/fake-s3 and change the header case to match what aws-sdk-java expects XD

hansonchar:
Or perhaps contributing patches to fake-s3 ?  After all, I suppose the whole point of fake-s3 is trying to, well, fake S3; so it seems reasonable to mimic the S3 peculiar behavior, even if it is not idealistically pure in terms of case sensitivity.

PR #437, opened 2015-05-29
case insensitive metadata header parsing

jackphel:
regarding #436


hansonchar:
Thanks for the contribution. The changes look reasonable.  Will do some further testings to verify.

hansonchar:
Hi @jackphel,

Turns out this patch causes many of our existing integration tests to fail, and evidently not safe.  For now, I'd suggest to make use of the work around suggested at:

  https://github.com/aws/aws-sdk-java/issues/436

PR #442, opened 2015-06-03
TransferManager: Resumable downloadDirectory

acmcelwee:
I ran into an unresolved issue of a SocketException causing `downloadDirectory` calls to fail in my `TransferManager` instances. @david-at-aws mentioned that the code explicitly does not retry on `SocketException`, so I've currently wrapped my downloadDirectory calls in retry logic, but retrying at that level means that you lose any progress you've made so far, and if you're dealing with large s3 "directories", that's potentially a lot of work to redo.

This updates the `doDownload` call in `downloadDirectory` to specify `resumeExistingDownload`.

### DIFF
--- a/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/transfer/TransferManager.java
+++ b/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/transfer/TransferManager.java
@@ -855,6 +855,37 @@ public class TransferManager {
      *            be created as necessary.
      */
     public MultipleFileDownload downloadDirectory(String bucketName, String keyPrefix, File destinationDirectory) {
+        return downloadDirectory(bucketName, keyPrefix, destinationDirectory, false);
+    }
+
+    /**
+     * Downloads all objects in the virtual directory designated by the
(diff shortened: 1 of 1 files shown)


david-at-aws:
Hey Adam,

I'm not sure we can safely change the default here - we might have people using this method and relying on the fact that it'll completely re-download the objects from S3, overwriting data in similarly-named files on the local disk instead of assuming the data in those files is a prefix of the object in S3. Can you make this configurable via TransferManagerConfiguration and/or an extra parameter to downloadDirectory that defaults to false so users can opt-in if it's safe for them to do so?

acmcelwee:
Doh! Not sure what I was thinking there. PR updated.

acmcelwee:
Looks like this is going to need more work. I pulled it into a project and tried it, but it throws an exception on files that have already been downloaded. I'll see if I have time over the next couple of days to improve on the PR.

david-at-aws:
Thanks Adam! New change looks good if we can sort out the error on existing files.

breedloj:
Closing this for now due to lack of activity.  Feel free to re-open if you've had a chance to update the PR.  Thanks!

PR #443, opened 2015-06-07
Fixed typo in the comment for putObject

craig-stevenson:
There was a typo in the comment for the put putObject function.  "this operation will" was written twice in a row.


manikandanrs:
Thanks for fixing this.

Issue #452, opened 2015-06-24
Amazon S3 Client doesn't retry on java.net.UnknownHostException


jakozaur:
Temporary network blip may cause DNS resolution to fail. Contrary to other network failures (IOException), AWS SDK doesn't retry by default on it.

Example failed stack trace:
```
com.amazonaws.AmazonClientException: Unable to execute HTTP request: stag-dual-custodian-store-000000000000141d.s3.amazonaws.com
	at com.amazonaws.http.AmazonHttpClient.executeHelper(AmazonHttpClient.java:472)
	at com.amazonaws.http.AmazonHttpClient.execute(AmazonHttpClient.java:296)
	at com.amazonaws.services.s3.AmazonS3Client.invoke(AmazonS3Client.java:3737)
	at com.amazonaws.services.s3.AmazonS3Client.putObject(AmazonS3Client.java:1462)
	at com.amazonaws.services.s3.transfer.internal.UploadCallable.uploadInOneChunk(UploadCallable.java:132)
	at com.amazonaws.services.s3.transfer.internal.UploadCallable.call(UploadCallable.java:124)
	at com.amazonaws.services.s3.transfer.internal.UploadMonitor.call(UploadMonitor.java:125)
	at com.amazonaws.services.s3.transfer.internal.UploadMonitor.call(UploadMonitor.java:47)
	at java.util.concurrent.FutureTask.run(FutureTask.java:262)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1145)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:615)
	at java.lang.Thread.run(Thread.java:745)
Caused by: java.net.UnknownHostException: stag-dual-custodian-store-000000000000141d.s3.amazonaws.com
	at java.net.InetAddress.getAllByName0(InetAddress.java:1250)
	at java.net.InetAddress.getAllByName(InetAddress.java:1162)
	at java.net.InetAddress.getAllByName(InetAddress.java:1098)
	at org.apache.http.impl.conn.SystemDefaultDnsResolver.resolve(SystemDefaultDnsResolver.java:45)
	at org.apache.http.impl.conn.DefaultClientConnectionOperator.resolveHostname(DefaultClientConnectionOperator.java:259)
	at org.apache.http.impl.conn.DefaultClientConnectionOperator.openConnection(DefaultClientConnectionOperator.java:159)
[...]

PR #453, opened 2015-06-24
Retry on DNS failure.

jakozaur:
Fixes our issue that we hit in production. Temporary network failures can cause DNS fail to resolve which is not retried. However, since AWS SDK is retrying on IOException. It should also retry on DNS failures.

Tests:
- [x] aws-sdk-java/aws-java-sdk-core master$ mvn test

Fixes:
https://github.com/aws/aws-sdk-java/issues/452


david-at-aws:
Hey Jacek, [UnknownHostException](http://docs.oracle.com/javase/8/docs/api/java/net/UnknownHostException.html) *is* an IOException, so it should already be getting retried via the ```if (exception.getCause() instanceof IOException)``` check above. Are you seeing cases where these retries aren't kicking in?

It's worth mentioning that Java's [default negative DNS cache ttl](http://docs.oracle.com/javase/8/docs/technotes/guides/net/properties.html) is 10 seconds, so unless you're tuning that down there's a fair chance that retrying will just hit the local DNS cache and fail again...

PR #459, opened 2015-07-06
Rewind ByteBuffer from MessageAttributeValue

jentfoo:
This automatically rewinds the ByteBuffer before returning.
SQS and SNS's MessageAttributeValue's behavior has been updated to remain consistent.

This seems to be a necessary behavior.  If the modification behavior is not desired then I recommend that the javadocs be updated to indicate to the developer that a rewind of the buffer may be needed after requesting it.

### DIFF
--- a/aws-java-sdk-sqs/src/main/java/com/amazonaws/services/sqs/MessageMD5ChecksumHandler.java
+++ b/aws-java-sdk-sqs/src/main/java/com/amazonaws/services/sqs/MessageMD5ChecksumHandler.java
@@ -278,12 +278,10 @@ public class MessageMD5ChecksumHandler extends AbstractRequestHandler {
      * input ByteBuffer and all the bytes it contains.
      */
     private static void updateLengthAndBytes(MessageDigest digest, ByteBuffer binaryValue) {
-        // Rewind the ByteBuffer, in case that get/put operations were applied to
-        // the unmarshalled BB before it's passed to this handler.
-        binaryValue.rewind();
-        int size = binaryValue.remaining();
+        ByteBuffer readOnlyBuffer = binaryValue.asReadOnlyBuffer();
+        int size = readOnlyBuffer.remaining();
         ByteBuffer lengthBytes = ByteBuffer.allocate(INTEGER_SIZE_IN_BYTES).putInt(size);
         digest.update(lengthBytes.array());
-        digest.update(binaryValue);
+        digest.update(readOnlyBuffer);
     }
 }


jentfoo:
It does not look like the travis CI tests failed due to a change I made?  In two of the runs it got:

java.lang.ClassNotFoundException: com.amazonaws.metrics.internal.cloudwatch.DefaultMetricCollectorFactory

Let me know if you think there needs to be anything I need to do to make this work.

shorea:
Could you provide some more context on what problems this is addressing?

jentfoo:
@shorea Absolutely...Basically these ByteBuffers are very likely (if not always) being consumed inside the AWS library.  The library user must implicitly know that they must rewind the buffer (ie they are not being provided a buffer which they can use directly).

As a library user when you are provided an object, you expect it to be usable immediately.  You don't expect to have to do an additional operation once it has been returned to you so that it can be usable (in this case that additional operation is a rewind).

If this PR is not accepted at a minimum the javadocs should be updated to inform the user that they must call .rewind() on the ByteBuffer once it is provided to them (or it may not be usable since the data was already consumed inside the AWS library).

shorea:
I agree this is very unintuitive. I wish we could take the change as is but unfortunately it is a semantic breaking change that could negatively impact customers relying on getBinaryValue to not modify the state of the ByteBuffer. We will consider making this change in our next major version bump as it does seem to be a more robust solution to this problem.

For now I would recommend changing places were we consume the ByteBuffer internally to use a read only view of that buffer. For SQS it looks like only MessageMD5ChecksumHandler#updateLengthAndBytes will modify the state of the ByteBuffer. SNS doesn't seem to have any code that would change the state of the ByteBuffer in MessageAttributeValue.

Thanks for bringing this to our attention! I'll look into what's going on with the Travis build to make sure your contribution can be merged.

jentfoo:
Awesome, thank you very much.  Let me know what I can do to help (I am not sure what is going on with travis.ci)

If you don't want to pull this in now I recommend expanding the javadocs to indicate that the ByteBuffer may have been consumed inside the AWS library, and {@link ByteBuffer#rewind()} should be invoked before it can be used.

shorea:
If you modify updateLengthAndBytes to use the ByteBuffer returned by binaryValue.asReadOnlyBuffer() instead of the original ByteBuffer (I think the rewind is still needed as the read only buffer will have the same position as the original, or at least that's the impression I get from the Javadocs) and revert all the accessor changes then we can accept this pull request. 

I've created an internal task for us to improve the documentation around ByteBuffer accessors.

shorea:
@jentfoo, did you want to make the recommended changes? I can drive this internally if you don't have the time.

jentfoo:
@shorea Sorry for going silent.  I was thinking it might make more sense to ensure the ByteBuffer is ready to consume from.  Then if any internal requests and usage does a .slice() before operating on the buffer, that would remove the need to rewind() the ByteBuffer returned from AWS UNLESS the library user consumes it, then request it again.

For that reason I still think rewinding it before returning makes the cleanest API.  But it would still be an improvement at least.  What do you think of that solution?

I have wanted to update this PR, but things kind of gotten kind of busy for me.  I still think I could get an update to you guys by Wed or Thursday likely.

jentfoo:
So here is the possible alternative I was talking about above.  This diff would be applied to master (not this branch):

https://gist.github.com/jentfoo/4042262f55362bebb276

I was surprised that I did not see any other cases where a provided ByteBuffer could be consumed (without making a readOnlyCopy first, which should be sufficent since position & limit are independent between the two buffers).  It also seemed like the initial buffer provided internally should always be ready to read from.

So in _theory_ this should be enough to fix my issue.  However I did not want to update the PR without showing you first (particularly since travis is failing).  Let me know what you think...If you want I can force push an update to this PR with the above change.

david-at-aws:
The new diff looks good to me. @shorea?

shorea:
Yep looks good to me. Go ahead and update the PR and I'll merge it in. Don't worry about the Travis build, it's having unrelated issues that I haven't had time to address.

jentfoo:
@shorea I just force pushed an update to this PR.

I should note that the risk introduced in my change is if I missed a condition where the buffer was being consumed (or if a new condition is added where it is consumed), it will no longer be rewound at this point.  Meaning that this MD5ChecksumHandler would be provided an already consumed ByteBuffer.

So it would be worth someone who knows more about this library to double check and ensure that they too don't see any points where internal to the library the ByteBuffer might be consumed (on a non-read only copy).

Still I feel pretty good about it from what I have seen.

shorea:
I didn't see any other places where it was being consumed internally. It should be safe to call .rewind() on the read only buffer as an extra precaution though since it has an independent position from the original ByteBuffer.

jentfoo:
> It should be safe to call .rewind()

Would you like me to add that?  I had considered that, but was unsure about it.  In theory it should not be _necessary_.  It also seems inconsistent (there does not appear to be any other places where we would rewind the buffer).  I could add it to just this point, or add it all points we get a read only version, or leave it how it is currently.  I am fine with any option since you are obviously more familiar with this code than me.

shorea:
I'm fine leaving this PR as is. It would only be necessary if we weren't sure if other parts of the code were incorrectly changing the state of the ByteBuffer instead of requesting a read only copy. Not rewinding it may help expose potential future issues where we are misusing the buffer internally since the MD5 validation would then fail.

jentfoo:
Cool, well it has been updated to master.  Let me know if there is anything else you want/need from me.

Thanks for all your guys help in getting this through!

shorea:
Thanks for the contribution @jentfoo. I'll get this merged in and staged for the next release (we've got one scheduled for later today but I'm not sure if it will make it in time so it may be next week). Can you quickly confirm you are submitting under the Apache 2.0 license?

jentfoo:
Sounds good.  And I can confirm, I have no issues with Apache 2.0 licensing.

shorea:
Good news, we are able to include this in today's release. Will be available in 1.10.8 version later today.

PR #471, opened 2015-07-25
Quote ETag while copying a part

andrewgaul:
This better matches the API documentation:

https://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectCOPY.html

### DIFF
--- a/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/model/transform/RequestXmlFactory.java
+++ b/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/model/transform/RequestXmlFactory.java
@@ -50,7 +50,7 @@ public class RequestXmlFactory {
             for (PartETag partEtag : partETags) {
                 xml.start("Part");
                 xml.start("PartNumber").value(Integer.toString(partEtag.getPartNumber())).end();
-                xml.start("ETag").value(partEtag.getETag()).end();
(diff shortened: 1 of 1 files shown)


david-at-aws:
Hey Andrew, are you seeing problems if the quotes are left out? If there's no benefit to including them (and as far as I can tell there's not?), I'd rather leave them out and save a couple bytes.

andrewgaul:
Not all implementations of the S3 API have loose ETag matching, e.g., [S3Proxy](https://github.com/andrewgaul/s3proxy).  Instead of prematurely optimizing out a couple of bytes shouldn't the implementation match the documentation?  Further all other uses have quoted ETags in the documentation and source.

breedloj:
While I do see them included with responses, I'm not seeing anything in the S3 API spec that states that a eTag should be surrounded by quotes as part of a request.  I'm not seeing any reason to change this unless there are specific cases where the existing implementation does not work.  Additionally I don't believe any other language AWS SDKs do this (.NET's implementation is here: https://github.com/aws/aws-sdk-net/blob/6c3be79bdafd5bfff1ab0bf5fec17abc66c7b516/sdk/src/Services/S3/Custom/Model/Internal/MarshallTransformations/CompleteMultipartUploadRequestMarshaller.cs#L66-L67).

andrewgaul:
Abandoning.

PR #473, opened 2015-07-29
Update region and service information for logs service.

duchatran:
Ref: http://docs.aws.amazon.com/general/latest/gr/rande.html#cw_region


david-at-aws:
Thanks for the contribution!

PR #479, opened 2015-08-05
Allow the specification of DnsResolver to use

tootedom:
Hi there,

This is a patch to allow the user to specify the DnsResolver to use to lookup ip addresses for AWS services.  For example, one could pass a DnsResolver that does exactly the same as the SystemDnsResolver:

    InetAddress.getAllByName(host)

But performs the request in the background so that foreground requests do not suffer from any slight "blips" in dns resolution times.  

thanks
/dom

### DIFF
--- a/aws-java-sdk-core/src/main/java/com/amazonaws/ClientConfiguration.java
+++ b/aws-java-sdk-core/src/main/java/com/amazonaws/ClientConfiguration.java
@@ -22,6 +22,9 @@ import com.amazonaws.http.IdleConnectionReaper;
 import com.amazonaws.retry.PredefinedRetryPolicies;
 import com.amazonaws.retry.RetryPolicy;
 import com.amazonaws.util.VersionInfoUtils;
+import org.apache.http.conn.DnsResolver;
+import org.apache.http.impl.conn.SystemDefaultDnsResolver;
+
 /**
  * Client configuration options such as proxy settings, user agent string, max
  * retry attempts, etc.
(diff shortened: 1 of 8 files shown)


bamdadd:
+1

david-at-aws:
I'm 100% on board for making this configurable, but I'm a bit nervous about exposing Apache HttpClient interfaces through our public API - we keep talking about making the HTTP client pluggable, which gets trickier if our public interface requires the use of AHC types. Would you be okay switching this to define a `DnsResolver` interface of our own and internally adapting it to Apache's `DnsResolver`?

tootedom:
Hi there,

Yup, makes complete sense.  

I'll create a DnsResolver Interface in `com.amazonaws` and in the `com.amazonaws.http.ConnectionManagerFactory` instantiate a `DelegatingDnsResolver` (which is an implementation of `org.apache.http.conn.DnsResolver`) which just takes in the constructor your com.amazonaws.DnsResolver implementation and calls it. 

Will send you/update this pull request in a couple of hours.
/dom

tootedom:
Hi there,

Apologies for the delay.  I've updated the pull request, and introduced a aws `DnsResolver` interface for the config

/dom

yurykats:
+1

tootedom:
Hi there,

Did you get any time to have a look at the pull request?  Do you require any changes to be made?

thanks
/dom

david-at-aws:
Hey, sorry for losing track of this. We were discussing whether we wanted to keep DnsResolver/SystemDefaultDnsResolver at the top level or stash them somewhere else out of the way, but couldn't come up with anywhere significantly better - top level's probably fine.

I'll run this past the lawyers (if you can preemptively confirm that you're contributing it under the Apache 2.0 license that'd be great :)) and hopefully get it merged in first thing next week.

tootedom:
>> (if you can preemptively confirm that you're contributing it under the Apache 2.0 license that'd be great :)

Yup I can confirm, that I'm contributing under the Apache 2.0 license.  
All Yours :-)

cheers
/dom

david-at-aws:
Good to go, thanks for the contribution (and especially for adding test coverage)!

PR #482, opened 2015-08-09
Correct typo on ALPHABETS

brymaven:
Correct typo on ALPHABETS on a few codec classes.


shorea:

PR #505, opened 2015-09-04
Change AWS4Signer newSigningKey method to protected

ejfinneran:
We'd like to keep credential secret keys in a separate secure system and
have that system generate the signingKey.

If we can override this method, that would make it easy to create a
subclass of AWS4Signer and use it as the signerOverride in the
ClientConfiguration


zhangzhx:
Hey @ejfinneran, this change looks good to me. I will forward this to our team for approval.

zhangzhx:
Sorry for the late merge. Can you confirm you are submitting under Apache 2.0?

ejfinneran:
Confirmed. Submitting this patch under Apache 2.0.

PR #512, opened 2015-09-18
SQS permissions apparently require that the 'sqs:' be capitalized

awebneck:
When attempting to use this class to construct and set SQS permissions on an existing queue, while the permissions save successfully, they don't seem to be functional. Using this to add the `SendMessage` permission to an SQS queue (for an SNS Topic ARN), for instance, results in the permission appearing to be set, but once the queue is subscribed to the SNS topic, messages sent to a queue with its permissions set via this method fail to show up therein, whereas messages sent to another queue which has had its permissions set via the AWS console appear as expected. The only difference between the two is capitalization, and resaving the first queue's permissions successfully changes the permission to a capitalized "SQS", whereafter the queue receives the messages as expected.

### DIFF
--- a/aws-java-sdk-sqs/src/main/java/com/amazonaws/auth/policy/actions/SQSActions.java
+++ b/aws-java-sdk-sqs/src/main/java/com/amazonaws/auth/policy/actions/SQSActions.java
@@ -21,43 +21,43 @@ import com.amazonaws.auth.policy.Action;
  */
 public enum SQSActions implements Action {
     /** Represents any action executed on Amazon SQS. */
-    AllSQSActions("sqs:*"),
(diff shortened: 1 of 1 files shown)


awebneck:
I should point out that this may simply be an internal issue with the interpretation of the permissions by SQS and/or SNS. That said, I have no channel to notify AWS about the problem if so, but this fix definitely solves it and will hopefully make them aware in the event the problem lies elsewhere. :)

awebneck:
Test failed due to PermGen space issue - not sure how this would have affected that. Trying to retrigger a fresh build.

awebneck:
Nevermind, it apparently doesn't even work when capitalized. Definitely a problem with the API, then.

Issue #510, opened 2015-09-15
SNS .util.SignatureChecker fails when it shouldn't due to bad parsing


acinader:
I'm creating an SNS message in the the AWS SNS ui.  My client is attempting to validate the raw json using  ```com.amazonaws.services.sns.util.SignatureChecker.verifyMessageSignature(String message, PublicKey publicKey)``` method.

It's failing with the following: 

```
java.lang.RuntimeException: Cannot process message of type String
	at com.amazonaws.services.sns.util.SignatureChecker.verifySignature(SignatureChecker.java:105)
```

The problem is that ```private Map<String, String> parseJSON(String jsonmessage)``` finds more than one ```fieldName``` of ```Type``` and it overwrites the correct Type, "Notification", with the incorrect Type, "String", that is found while parsing the ```MessageAttributes``` section.

Below is a sample message that will trip the bug.

I'd suggest fixing by having a POJO for a message that jackson maps into.  Not sure what the pull request policy is, but I can cook up if it isn't a waste of time.

the message 

```
{
  "Type" : "Notification",
  "MessageId" : "85c8859f-4201-5b71-ab61-cbed06183f8e",
  "TopicArn" : "arn:aws:sns:us-east-1:810605503585:just_testing",
  "Subject" : "another test",
  "Message" : "this is a test",
  "Timestamp" : "2015-09-15T16:05:45.300Z",
  "SignatureVersion" : "1",
  "Signature" : "UH6scc3l4MjKtfJ+0gcrEW+/IxMfJ05KG9G9wCKIyEkTV+W5tLBNrsp9uvnqHxX2OGCaVFqEJgYwmHa5O65tyUvtRJRUbo/fFSZKI3GHQUWd8RZ4ro0NYyE/DQTkBufKG406mrmZRndS9ajckLRgFF4gbL7bbHas+GGdONsSHylvV2fuOUuiTnlQk87N1hDTsQa56Mic/eFF4BEylZ5tqr5qNvEDctR+2BkWlL+fUwfU6PSfiJBJFIlYABr+tSLHmk0xqkg5Ue7mlmoyvFPLMMf/SKoSJo8uLXrkie2pT0b/eIV3RfauvD7vwYwZhUTDOQSxXovV3u/P2ZA0z/T0Sg==",
  "SigningCertURL" : "https://sns.us-east-1.amazonaws.com/SimpleNotificationService-bb750dd426d95ee9390147a5624348ee.pem",
[...]

PR #513, opened 2015-09-20
Only parse fields we need to validate the message

acinader:
The simple message parser was parsing MessageAttributes that was causing the important Type
field to get overwritten incorrectly.  By ignoring fields that we don't need, we wont overwrite
top level fields with nested fields.

#510


shorea:
Change looks good, running through some internal tests before merging. Can you confirm you are submitting under the Apache 2.0 license?

acinader:
@shorea yes, submitted under apache 2.0 and I added the license to the new test file.

PR #517, opened 2015-09-23
Added test for StringUtils.replace and fix for bug.

alexmojaki:
Wanted to use replace for escaping and realised there was a bug that can lead to an infinite loop. Will add a fix next.

### DIFF
--- a/aws-java-sdk-core/src/main/java/com/amazonaws/util/StringUtils.java
+++ b/aws-java-sdk-core/src/main/java/com/amazonaws/util/StringUtils.java
@@ -141,7 +141,7 @@ public class StringUtils {
         int indexOf = buffer.indexOf( partToMatch );
         while ( indexOf != -1 ) {
             buffer = buffer.replace( indexOf, indexOf + partToMatch.length(), replacement );
-            indexOf = buffer.indexOf( partToMatch );
+            indexOf = buffer.indexOf( partToMatch, indexOf + replacement.length() );
         }
 
         return buffer.toString();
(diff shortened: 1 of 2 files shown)


alexmojaki:
I was hoping to have the build fail after I added the test to demonstrate the need for the fix, but it didn't trigger a build.

shorea:
Thanks for the contribution! Can you confirm you are submitting under the Apache 2.0 license? Change looks good and kudos for the test case. Just wanted to get a sense of how you encountered this bug. Are you using StringUtils directly? If so what advantage does it provide over String.replaceAll?

alexmojaki:
I'm not sure how to confirm that I am submitting under a given license, I've never heard of such a request. I have no objection to my changes being used under the license.
I wanted to do a string replace to efficiently create a properly escaped CSV by first replacing all `\` with `\\` and then all `,` with `\,`. String.replaceAll is very slow as it compiles a new Pattern object every time it is called. You can precompile the regex but there's still overhead as it's an overly complicated method. I typed in `StringUtils.<ctrl+space>` to autocomplete not really knowing what I would get, kind of expecting something from Apache Commons. Looked at the source code out of interest. Made a custom version that did multiple replaces in sequence so as to only create one StringBuilder. Later, but before having actually tried it out, I was randomly thinking about it and realised the bug. Tested and confirmed.

shorea:
"I have no objection to my changes being used under the license." This is confirmation enough. Our internal open source team just needs something explicit to state the customer intends to submit under the license we use.

Thanks again for the contribution. I'll get this approved and merged in.

PR #533, opened 2015-10-12
Make session renewal atomic in STSAssumeRoleSessionCredentialsProvider

nezihyigitbasi:
Currently if multiple threads share the same
`STSAssumeRoleSessionCredentialsProvider` instance, on a `getCredentials()`
call if the credentials are expired, multiple threads may get a true when
they call `needsNewSession()`, and hence multiple threads may `startSession()` causing
a surge in assume role calls. This commit fixes this issue by making
session renewal atomic.

### DIFF
--- a/aws-java-sdk-sts/src/main/java/com/amazonaws/auth/STSAssumeRoleSessionCredentialsProvider.java
+++ b/aws-java-sdk-sts/src/main/java/com/amazonaws/auth/STSAssumeRoleSessionCredentialsProvider.java
@@ -16,6 +16,8 @@
 package com.amazonaws.auth;
 
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
 
 import com.amazonaws.ClientConfiguration;
 import com.amazonaws.annotation.ThreadSafe;
@@ -57,6 +59,10 @@ public class STSAssumeRoleSessionCredentialsProvider implements AWSSessionCreden
     
     /** The Duration for assume role sessions. */
     private final int roleSessionDurationSeconds;
+
+    /** Lock used to serialize access to expiration check and session renewal logic across
+     * multiple threads sharing this instance */
+    private final Lock sessionRenewalLock = new ReentrantLock();
 
     /**
      * Constructs a new STSAssumeRoleSessionCredentialsProvider, which makes a
@@ -243,8 +249,13 @@ public class STSAssumeRoleSessionCredentialsProvider implements AWSSessionCreden
     
     @Override
     public AWSSessionCredentials getCredentials() {
-        if (needsNewSession()) {
-            startSession();
+        try {
+            sessionRenewalLock.lock();
+            if (needsNewSession()) {
+                startSession();
+            }
+        } finally {
+            sessionRenewalLock.unlock();
         }
         Credentials tempCredentials = sessionCredentials.get();
         return new BasicSessionCredentials(tempCredentials.getAccessKeyId(),


shorea:
Hey @nezihyigitbasi, thanks for the pull request. Are you running into specific problems with multiple calls to STS being made? This issue came up in the code review for this change and we explicitly choose not to lock because a bad connection or socket could cause other calls to getCredentials to be blocked a long time. Several options were proposed but in the end we went with the simple approach (option 1 below) as it was still an improvement over the original non thread-safe code. I've included the correspondence in the code review below. 

```

1. Get back to a previous version without lock, but with possible unnecessary STS calls:

if (needsNewSession()) {
    startSession();
}

Pros: No locks during service call
Cons: Possible unnecessary STS calls within expiration time

2. Instead of using synchronized(object){}, use Lock.tryLock(timeout), but in that case the proper timeout value is unclear, because it should depend on ClientConfiguration parameters, i.e. AmazonWebServiceClient configuration.

if (needsNewSession()) {
    if(lock.tryLock(timeout)) {
        try {
            if(needsNewSession()) {
                startSession();
            }
        }
        finally {
            lock.unlock();
        }
    }
    else {
        if(needsNewSession()) {
            // throw some exception()
        }
    }
}

Pros:
* No indefinite locks during service call
Cons:
* Unclear timeout value
* User code will get some vague exception in case of failure to lock by timeout.

3. Use ScheduledExecutorService to call startSession() every, for example, (roleSessionDurationSeconds - 2 * EXPIRY_TIME_MILLIS), and use the previous version of needsNewSession() without a lock. In that case ScheduledExecutorService will update credentials in timely manner and requesting threads won't need to call STS by themselves. The thread running by ScheduledExecutorService will swallow all exceptions and if this happened, then the provider will fall back to updating credentials by user threads. If the issue with updating credentials remains, then the user will get correct exception from STS from user threads.

private final ScheduledExecutorService scheduler = Executors.newScheduledThreadPool(1);

...

private STSAssumeRoleSessionCredentialsProvider(Builder builder) {
    ...
    
    Runnable sessionRefresher = new Runnable() {
        public void run() {
            try {
                startSession();
            }
            catch (Exception ex) {
                // do nothing
            }
        }
    };
    
    scheduler.scheduleAtFixedRate(sessionRefresher, 0, roleSessionDurationSeconds - 2 * EXPIRY_TIME_MILLIS, SECONDS);
}

...

if (needsNewSession()) {
    // A user thread gets to this point if last time sessionRefresher caught an exception.
    // If the issue which prevented sessionRefresher from updating credentials remains, then the current thread will get the same exception and user code will be notified
    startSession();
}

Pros:
* No locks during service call
* User threads don't waste time calling STS
Cons:
* Possible unnecessary STS calls by user threads in case of failure of sessionRefresher
```

nezihyigitbasi:
@shorea The issue that we ran into is excessive assume role calls that cause our calls to be throttled. Then we get alerts/emails from the aws team and we spend time investigating the root cause of that.

>> We explicitly choose not to lock because a bad connection or socket could cause other calls to getCredentials to be blocked a long time.

Even without locking if the connection or socket is bad the threads getting credentials will block for long time too (all threads will make network calls and block). With locking we get rid of multiple assume role calls, but the blocking time will roughly be the same. won't it? AFAIK there are timeout values for network calls in the sdk, so blocking is not infinite anyway. Once the connection becomes good again, threads will start making progress.

shorea:
The concern is one bad call (slow because of a hung connection or just slow server side) can cause many threads to be blocked waiting on STS Credentials. We've also seen cases in buggy version of the Apache HTTP client where socket timeout is ignored and connections hang forever. 

If this is causing issues for you though we can look into other options. I think a modified version of option two could be interesting. One of the cons listed was an unclear timeout value, the timeout doesn't necessarily need to come from the client configuration though. It just needs to be set generously enough to avoid throttling exceptions in STS. Also if threads are unable to acquire the lock they can still return the current credentials as long as they are still valid (we try and refresh a minute before the creds expire).

nezihyigitbasi:
Thanks @shorea for the detailed replies. But I am still not convinced that a bad network condition affects the code with the lock (what this PR proposes) more than the current code (without the lock). 

For a particular `STSAssumeRoleSessionCredentialsProvider` instance there are two cases:
1) **Credentials have not expired yet:** For the current code all threads will get new `BasicSessionCredentials` instances, for the code with the lock threads will just acquire the lock one by one and will see that `needsNewSession()` is false and then simply get new `BasicSessionCredentials` instances. So no network calls at all, only overhead of locking.
2) **Credentials have expired:** For the current code all threads will see that `needsNewSession()` is true and they will all make network calls to refresh the credentials and they will all get stuck, say for T seconds and in total will make N assume role calls. For the code with the lock one thread will acquire the lock and make a network call and get stuck for T seconds, while the other threads will get stuck holding the lock. After T seconds rest of the threads will just see that  `needsNewSession()` is false and move on, so the the total time that the threads get stuck will be roughly T seconds and in total only 1 assume role calls will be made. So both with and without locking the wait time will be roughly T seconds.

Am I missing something here? It seems like there shouldn't be much difference between the two implementations except with the lock we make less number of assume role calls and we introduce some locking overhead, which I think won't be a big issue.

nezihyigitbasi:
BTW my previous comment holds if the lock is not static.

shorea:
There is also a third case where credentials are about to expire. If the creds are within a minute of expiration time then needsNewSession will return true and we fetch new ones from the service. Subsequent calls to getCredentials may not need to be blocked waiting for the call to succeed as the credentials could still be valid. That leads to some debate on when we should stop serving credentials that are about to expire as it doesn't make much sense to return credentials that will expire before you can do anything useful with them.

I think the main concern here is that one slow call will become a bottleneck for all threads attempting to get credentials when they need to be refreshed. The current solution seems less than ideal as it causes a flood of calls to be made and makes no attempt to limit them. The solution you've proposed removes the throttling concern but introduces some potentially unwanted performance impacts if all threads are blocked on a slow or non-responsive network call. I agree with you that the current solution is worse then what you have proposed as it can cause credentials to never be available or can impact code outside the credentials provider due to the throttling of STS.  We need something in the middle where we don't flood STS with requests but we also don't wait too long on one particular service call and kick off another instead. 

Option two I discussed before has some merits. I was also chatting with a colleague and he proposed an interesting solution I rather like. It's basically a version of option 3 where we use a executor to automatically refresh credentials in the background once the expiration is within the threshold. The thread pool itself has some keep alive time and calls to getCredentials refresh that keep alive time. So if the credentials provider stops being used but is still in memory then we want continue to regularly call STS until the JVM terminates, it will eventually be shut down and revived if the provider starts to be used again.

paulfrmbstn:
How about a lock-free solution? https://github.com/aws/aws-sdk-java/pull/570

shorea:
I'm a little more in favor of a more asynchronous approach. With the suggested approach one thread will still be locked refreshing credentials. I think it would be ideal if we just submit a task on our own internal thread rather then blocking the customers. I think we need to make a better distinction between these credentials are expiring and need to be refreshed versus these credentials are expired and I can't return until I get some new credentials.

This is a little off topic but it would be cool to have a method that would take in the time you plan to use the credentials and from that determine if we can use the current cache or have to call out to the service to get new ones. This seems a little more flexible then the second second hard coded limit we have now.

paulfrmbstn:
I updated https://github.com/aws/aws-sdk-java/pull/570 with an async refresh. Take a look and let me know if that is closer to what you are thinking. I think I worked through most if not all of the edge cases.

shorea:
Closing in favor of #570

paulfrmbstn:
#570 is for STSSessionCredentialsProvider, does it not make sense to still track  STSAssumeRoleSessionCredentialsProvider here? Or do you want #570 to apply the same solution to STSAssumeRoleSessionCredentialsProvider?

shorea:
Good call. Didn't realize they were for different credentials provider. If you are willing to open another pull request for assume role provider with a similar implementation I can review it.

paulfrmbstn:
Those two classes are very similar. Should they extend a common abstract base class with the common implementation or kept as distinct implementations of AWSSessionCredentialsProvider?

shorea:
Yeah that sounds good. I believe the only thing that should differ are the actual API calls to STS.

PR #545, opened 2015-10-22
Use Content-Type application/octet-stream when not set

andrewgaul:
Previously the SDK provided application/x-www-form-urlencoded, which
confused servers like S3Proxy.  The underlying Jetty interpreted a
multi-part put object entity body as form data, exhausting the fixed
buffer size.  Fixes andrewgaul/s3proxy#80.


ankon:
Instead of `application/unknown`, what about using `application/octet-stream`, which is simply "bunch of bytes", and understood by applications as such?

See also the RFC quotes in http://stackoverflow.com/questions/1176022/unknown-file-type-mime/28652339#28652339 (and the links there)

shorea:
I like the octet stream suggestion, ran our integration tests last night and no regressions by changing the default to application/octet-stream. Can you update the PR and confirm you're submitting under apache 2.0?

andrewgaul:
I updated the PR with application/octet-stream and confirm that I submit this change under the Apache 2.0 license.

shorea:
Thanks! I've staged this internally, just waiting for approval from legal and I'll get this merged in.

shorea:

PR #560, opened 2015-11-19
Add property to disable Put MD5 validation

andrewgaul:
Some object stores like S3Proxy with the Azure backend do not return
the MD5 sum for ETag.  Fixes andrewgaul/s3proxy#96.


andrewgaul:
We could also change the property to be more generic since object stores which do not return MD5 ETag for put also do not return it for get or copy.

shorea:
Change looks good to me. Can you confirm you are submitting under the apache 2.0 license?

andrewgaul:
Confirmed that I submit this change under the Apache 2.0 license.

shorea:
Thanks! I've staged this internally, just waiting for approval from legal and I'll get this merged in.

shorea:

PR #570, opened 2015-12-07
A lock-free solution for the throttling issue in STSSessionCredentialsProvider

paulfrmbstn:
High throughput applications are nearly guaranteed to get STS throttling errors when creating or refreshing temporary credentials when multiple threads are using the same STSSessionCredentialsProvider instance.

### DIFF
--- a/aws-java-sdk-sts/src/main/java/com/amazonaws/auth/STSSessionCredentialsProvider.java
+++ b/aws-java-sdk-sts/src/main/java/com/amazonaws/auth/STSSessionCredentialsProvider.java
@@ -15,33 +15,89 @@
 
 package com.amazonaws.auth;
 
-import java.util.Date;
-
+import com.amazonaws.AmazonClientException;
 import com.amazonaws.ClientConfiguration;
+import com.amazonaws.annotation.ThreadSafe;
 import com.amazonaws.services.securitytoken.AWSSecurityTokenService;
 import com.amazonaws.services.securitytoken.AWSSecurityTokenServiceClient;
 import com.amazonaws.services.securitytoken.model.Credentials;
 import com.amazonaws.services.securitytoken.model.GetSessionTokenRequest;
 import com.amazonaws.services.securitytoken.model.GetSessionTokenResult;
+import java.util.Date;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.concurrent.locks.Condition;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
 
 /**
  * AWSCredentialsProvider implementation that uses the AWS Security Token
  * Service to create temporary, short-lived sessions to use for authentication.
  */
+@ThreadSafe
 public class STSSessionCredentialsProvider implements AWSSessionCredentialsProvider {
 
     /** Default duration for started sessions */
     public static final int DEFAULT_DURATION_SECONDS = 3600;
+    
+    /** Time before expiry within which credentials will be renewed. */
+    private static final int EXPIRY_TIME_MILLIS = 60 * 1000;
 
+    /** maximum time to wait for STS service to respond */
+    private static final long NEW_SESSION_MAX_WAIT_SECONDS = 5;
+    
+    /**
+     * Private class used to atomically store a session
+     * with its expiration time.
+     */
+    private static final class SessionCredentialsHolder {
+        private final AWSSessionCredentials sessionCredentials;
+        private final Date sessionCredentialsExpiration;
+        
+        private SessionCredentialsHolder(AWSSessionCredentials sessionCredentials,
+                                         Date sessionCredentialsExpiration) {
+            this.sessionCredentials = sessionCredentials;
+            this.sessionCredentialsExpiration = sessionCredentialsExpiration;
+        }
+        
+        @Override
+        public int hashCode() {
+            return sessionCredentials.hashCode();
+        }
+
+        @Override
+        public boolean equals(Object obj) {
+            if (null == obj) {
+                return false;
+            }
+            if (getClass() != obj.getClass()) {
+                return false;
+            }
+            final SessionCredentialsHolder other = (SessionCredentialsHolder) obj;
+            return !(this.sessionCredentials != other.sessionCredentials
+                     && (null == this.sessionCredentials || !this.sessionCredentials.equals(other.sessionCredentials)));
+        }
+    }
+    
     /** The client for starting STS sessions */
     private final AWSSecurityTokenService securityTokenService;
+    
+    /** Used to synchronize session creation */
+    private final Lock newSessionLock = new ReentrantLock();
+    private final Condition newSessionCondition = newSessionLock.newCondition();
 
-    /** The current session credentials */
-    private AWSSessionCredentials sessionCredentials;
+    /** The current session credentials and expiration time */
+    private final AtomicReference<SessionCredentialsHolder> sessionCredentials = new AtomicReference<SessionCredentialsHolder>();
+    
+    /** The last exception thrown during an asynchronous session refresh */
+    private final AtomicReference<Exception> lastException = new AtomicReference<Exception>();
+    
+    /** Used to interrupt the latest refresh thread */
+    private final AtomicReference<Thread> refreshThread = new AtomicReference<Thread>();
 
-    /** The expiration time for the current session credentials */
-    private Date sessionCredentialsExpiration;
-
+    /** Used to ensure only one thread at any given time refreshes the re-usable session */
+    private final AtomicBoolean asyncRefreshing = new AtomicBoolean(false);
 
     /**
      * Constructs a new STSSessionCredentialsProvider, which will use the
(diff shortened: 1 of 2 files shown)


shorea:
Commented on https://github.com/aws/aws-sdk-java/pull/533 to keep the conversation there.

shorea:
Thanks for updating. Reviewing now.

aws-java-sdk-sts/src/main/java/com/amazonaws/auth/STSSessionCredentialsProvider.java
```diff
@@ -35,13 +39,50 @@
 
     /** The client for starting STS sessions */
     private final AWSSecurityTokenService securityTokenService;
+    
+    /**
+     * Private class used to atomically store a session
+     * with its expiration time.
+     */
+    private static class SessionCredentialsHolder {
```
shorea: Can we make this class final and immutable?

aws-java-sdk-sts/src/main/java/com/amazonaws/auth/STSSessionCredentialsProvider.java
```diff
@@ -119,36 +160,119 @@ public STSSessionCredentialsProvider(AWSCredentialsProvider longLivedCredentials
...
         if (needsNewSession()) startSession();
 
-        return sessionCredentials;
+        return sessionCredentials.get().sessionCredentials;
     }
 
     @Override
     public void refresh() {
-        startSession();
+        asyncNewSession();
```
shorea: I think we should keep this synchronized.

aws-java-sdk-sts/src/main/java/com/amazonaws/auth/STSSessionCredentialsProvider.java
```diff
@@ -119,36 +160,119 @@ public STSSessionCredentialsProvider(AWSCredentialsProvider longLivedCredentials
...
     }
+    
+    /**
+     * Atomically create a new session in a lock-free manner. If an exception
+     * is not thrown, it means exactly one thread successfully set the new
+     * current session, but not necessarily the calling thread
+     */
+    private void casNewSession() {
```
shorea: What does cas mean?

aws-java-sdk-sts/src/main/java/com/amazonaws/auth/STSSessionCredentialsProvider.java
```diff
@@ -119,36 +160,119 @@ public STSSessionCredentialsProvider(AWSCredentialsProvider longLivedCredentials
...
+        SessionCredentialsHolder currentSession = sessionCredentials.get();
+
+        if (null == currentSession || expiring(currentSession.sessionCredentialsExpiration)) {
```
shorea: Can you use needsNewSession here?

aws-java-sdk-sts/src/main/java/com/amazonaws/auth/STSSessionCredentialsProvider.java
```diff
@@ -119,36 +160,119 @@ public STSSessionCredentialsProvider(AWSCredentialsProvider longLivedCredentials
...
+        }
+    }
+    
+    /**
+     * Used when there is no valid session to return. Only called
+     * for initial or expired session. Callers are blocked until
+     * exactly one new session is created or an exception is thrown.
+     */
+    private synchronized void serialzedNewSession() {
+        while (expired()) {
```
shorea: I don't think we should loop indefinitely here. If there is a problem retrieving credentials from STS any thread call getCredentials could hang forever. How about instead let's try and acquire the refreshing lock and give up after a certain time (say five seconds). I think this will cover 99.99% of cases where the first request succeeds in a timely fashion but doesn't hang forever in the .01% chance that the there is a problem. Even with this approach there is a possibility that we will hammer STS with calls once the timeout expires but I think that's the exceptional case and not the norm.

aws-java-sdk-sts/src/main/java/com/amazonaws/auth/STSSessionCredentialsProvider.java
```diff
@@ -119,36 +160,119 @@ public STSSessionCredentialsProvider(AWSCredentialsProvider longLivedCredentials
...
+    private void asyncNewSession() {
+        // immediate return if refresh already in progress
+        if (refreshing.compareAndSet(false, true)) {
+            try {
+                // current thread is chosen as the session refresh thread
+                new Thread() {
+                    @Override
+                    public void run() {
+                        // thread exits immediately if an async refresh already in progress
+                        if (asyncRefreshing.compareAndSet(false, true)) {
```
shorea: Can we avoid creating the thread altogether if asyncRefreshing is false?

aws-java-sdk-sts/src/main/java/com/amazonaws/auth/STSSessionCredentialsProvider.java
```diff
@@ -119,36 +160,119 @@ public STSSessionCredentialsProvider(AWSCredentialsProvider longLivedCredentials
...
+    /**
+     * Used when there is a current session to return. Ensures only one
+     * thread is spawned at any given time to refresh the session. Caller
+     * is never blocked.
+     */
+    private void asyncNewSession() {
+        // immediate return if refresh already in progress
+        if (refreshing.compareAndSet(false, true)) {
+            try {
+                // current thread is chosen as the session refresh thread
```
shorea: This comment is no longer applicable.

shorea:
Wondering if you've had a chance to look at my comments.

aws-java-sdk-sts/src/main/java/com/amazonaws/auth/STSSessionCredentialsProvider.java
paulfrmbstn: https://en.wikipedia.org/wiki/Compare-and-swap

aws-java-sdk-sts/src/main/java/com/amazonaws/auth/STSSessionCredentialsProvider.java
paulfrmbstn: asyncNewSession() is thread-safe and lock-free, multiple threads can concurrently call refresh() without side effects. All synchronized would do here is prevent liveness.

shorea: Sorry I meant synchronous not synchronized. My concern is that by doing the refresh asynchronously we are violating the contract of the interface. Refresh is more forceful and as a developer I'd expect the refresh to be completed when this method returns and not just queued up to be refreshed at some later time.

aws-java-sdk-sts/src/main/java/com/amazonaws/auth/STSSessionCredentialsProvider.java
paulfrmbstn: yes, I will refactor this a little

aws-java-sdk-sts/src/main/java/com/amazonaws/auth/STSSessionCredentialsProvider.java
paulfrmbstn: Yes, I will remove the builder setters in favor of a constructor.

aws-java-sdk-sts/src/main/java/com/amazonaws/auth/STSSessionCredentialsProvider.java
paulfrmbstn: Agree, this needs some sort of back-off here. Will do.

aws-java-sdk-sts/src/main/java/com/amazonaws/auth/STSSessionCredentialsProvider.java
paulfrmbstn: Yes. Will do.

aws-java-sdk-sts/src/main/java/com/amazonaws/auth/STSSessionCredentialsProvider.java
paulfrmbstn: true.

aws-java-sdk-sts/src/main/java/com/amazonaws/auth/STSSessionCredentialsProvider.java
paulfrmbstn: The interface contract is ambiguous in my opinion. To me, it is not strongly-worded and suggests the method is implementation-specific. But your point is valid, I can make it a blocking call.
/**
     * Forces this credentials provider to refresh its credentials. For many
     * implementations of credentials provider, this method may simply be a
     * no-op, such as any credentials provider implementation that vends
     * static/non-changing credentials. For other implementations that vend
     * different credentials through out their lifetime, this method should
     * force the credentials provider to refresh its credentials.
     */

shorea:
Just wanted to confirm this is good to review again? Github doesn't give any notifications when the diff is updated but it looks like the comments have been addressed.

paulfrmbstn:
I did address the comments, but not all that happy with the newSessionLock logic as it prevents all threads with an expired session from getting a new session when a current refresh request is experiencing a transient network/endpoint issue that will eventually time-out (but subsequently succeed). I wanted to put some logic in there that allows for tightly controlled secondary STS requests if the primary refresh is taking longer than some threshold. I have not had the time to implement that though. But, to answer your question, yes, I would be curious of your thoughts at this stage.

shorea:
Great! I'll take another look today.

aws-java-sdk-sts/src/main/java/com/amazonaws/auth/STSSessionCredentialsProvider.java
```diff
@@ -116,39 +167,186 @@ public STSSessionCredentialsProvider(AWSCredentialsProvider longLivedCredentials
...
+    private void asyncNewSession() {
+        // immediate return if refresh already in progress
+        if (refreshing.compareAndSet(false, true)) {
+            try {
+                if (false == asyncRefreshing.get()) {
+                    new Thread() {
+                        @Override
+                        public void run() {
+                            // thread exits immediately if an async refresh already in progress
+                            if (asyncRefreshing.compareAndSet(false, true)) {
```
shorea: Can we just move the compare and set outside the thread instead of having 
```java 
if(false == asyncRefreshing.get())
```

aws-java-sdk-sts/src/main/java/com/amazonaws/auth/STSSessionCredentialsProvider.java
```diff
@@ -116,39 +167,186 @@ public STSSessionCredentialsProvider(AWSCredentialsProvider longLivedCredentials
...
+
+            // wait for new session or time out
+            if (false == newSessionCondition.await(NEW_SESSION_MAX_WAIT_SECONDS, TimeUnit.SECONDS)) {
+                // timeout occured, check for valid session before throwing an exception
+                if (expired()) {
+                    throw new AmazonClientException("Session credentials request timed-out.", lastException.get());
+                }
+            }
+        } catch (InterruptedException ex) {
+            throw new AmazonClientException(ex);
```
shorea: Can you reinterrupt the thread so the caller can act on it?

aws-java-sdk-sts/src/main/java/com/amazonaws/auth/STSSessionCredentialsProvider.java
```diff
@@ -116,39 +167,186 @@ public STSSessionCredentialsProvider(AWSCredentialsProvider longLivedCredentials
...
+    
+    /**
+     * Session credentials that expire in less than a minute are considered expiring.
+     * 
+     * @param expiry expiration time of a session
+     * @return 
+     */
+    private static boolean expiring(Date expiry) {
+        long timeRemaining = expiry.getTime() - System.currentTimeMillis();
+        return timeRemaining < (60 * 1000);
```
shorea: Can we make the expring time a constant?

shorea:
Had just a couple more minor comments but other than that I think this is ready to merge. Can you confirm in the comments that you are submitting under the Apache 2.0 license.

aws-java-sdk-sts/src/main/java/com/amazonaws/auth/STSSessionCredentialsProvider.java
paulfrmbstn: My concern here is what happens if ```new``` or ```Thread.start()``` throws an exception. I prefer that to be at least potentially recoverable. If the campareAndSet is moved out of the thread, it would require intricate exception handling with maintainability issues. However, since ```refreshing``` is no longer being used external to ```asyncNewSession```, I think this logic can be simplified.

aws-java-sdk-sts/src/main/java/com/amazonaws/auth/STSSessionCredentialsProvider.java
paulfrmbstn: I think calling interrupt on the thread here is not a bad idea since the apache http client does check the interrupted flag when it can. That could be a poor-man's solution to the transient socket timeout issue I was thinking about solving in a more deterministic fashion. Will do the former for now.

aws-java-sdk-sts/src/main/java/com/amazonaws/auth/STSSessionCredentialsProvider.java
paulfrmbstn: Yes.

paulfrmbstn:
Latest comments addressed. Code is submitted under the Apache 2.0 license.

shorea:
Sorry for the delay on this, took a while to get through internal review. Some changes that were made per review comments.

1. blockingNewSession behaves a bit differently. It no longer wait on a condition triggered by the async refresh thread, it just tries to acquire a lock within the timeout and if the credentials are still expired when lock is acquired then it calls STS synchronously. If it can't acquire the lock within the timeout it just calls STS synchronously which has the a chance of causing throttling but this should be the exceptional case not the norm.
2. We use blockingNewSession when credentials are expired or expiring (within a minute) just in case people are relying on the existing behavior. Async refresh is now triggered within five minutes of expiration so hopefully blockingNewSession is very rarely called.

paulfrmbstn:
I see this merge was wiped out by 1.10.51 on master. Are the mentioned changes staged somewhere for the next release? I would like to use the new code as a basis for a similar PR for STSAssumeRoleSessionCredentialsProvider.

manikandanrs:
@paulfrmbstn 
Sorry for the wiping out on 1.10.51. This is getting released as part of today's release.

PR #559, opened 2015-11-18
Added type overrides to the cargs in decider registration

nterry:
This allows consumers to use polymorphism in constructor
resolution, rather than the literal built-in type resolution

This is needed to overcome a Java limitation

PR #572, opened 2015-12-08
Improved type resolution in decider registration cargs

nterry:
Brought in java.beans.Expression class to perform dynamic
constructor resolution. This is in response to some feedback
from Luc on the SWF team.

Also removed some redundant code and added a stub POJO to replace
an anonymous SWF POJO implementation.


nterry:
Previous PR is #559

nterry:
@shorea Thanks!!

PR #576, opened 2015-12-11
slight fix to wording in ClientConfiguration documentation

EronHennessey:
A couple of methods set timeouts but claimed to return them. This has been fixed.


shorea:
Thanks Eron!

PR #578, opened 2015-12-14
Updated broken URL for Jackson documentation

EronHennessey:
The existing URL pointed to a page that no longer exists (it looks like codehaus is no more.. https://www.codehaus.org/termination/).

I've updated the link to point to the current Jackson processor documentation, at http://wiki.fasterxml.com/JacksonHome

Thanks!


shorea:
Thanks Eron!

PR #581, opened 2015-12-18
Add an option to disable chunked encoding.

### LINKED ISSUE #580: Is there a way to not use chunked uploading for java-sdk for signature V4?
```PUT /testbbucket/caddy HTTP/1.1 1 1 map[X-Amz-Content-Sha256:[STREAMING-AWS4-HMAC-SHA256-PAYLOAD] User-Agent:[aws-sdk-java/1.10.41 Mac_OS_X/10.11.1 Java_HotSpot(TM)_64-Bit_Server_VM/25.25-b02/1.8.0_25] Content-Type:[application/octet-stream] Content-Length:[12241473] Connection:[Keep-Alive] X-Amz-Date:[20151218T164446Z] X-Amz-Decoded-Content-Length:[12232928] Content-Md5:[4uUkMUjnvb+ymFcDSeexUA==] Authorization:[AWS4-HMAC-SHA256 Credential=WLGDGYAQYIGI833EV05A/20151218/us-east-1/s3/aws4_request, SignedHeaders=content-length;content-md5;content-type;host;user-agent;x-amz-content-sha256;x-amz-date;x-amz-decoded-content-length, Signature=6575c62db878f4151357f54b4e7a5c19555afb1a9efc0573dce615850caaf646]] 0xc82017a140 12241473 [] false localhost:9000 map[] map[] <nil> map[] 127.0.0.1:50544 /testbbucket/caddy <nil> <nil>```

Following code makes it default - i couldn't find a way to override this, is there any? 

```java
/**
     * Determine whether to use aws-chunked for signing
     */
    private static boolean useChunkEncoding(SignableRequest<?> request) {
        // Whether to use chunked encoding for signing the request
        boolean chunkedEncodingEnabled = false;
        if (request.getOriginalRequestObject() instanceof PutObjectRequest
                || request.getOriginalRequestObject() instanceof UploadPartRequest) {
            chunkedEncodingEnabled = true;
        }
        return chunkedEncodingEnabled;
    }
```

The behavior is entirely different than how aws-sdk-go behaves which does calculate the sha256 during multipart and regular put operation.

harshavardhana:
This is done to allow sha256 calculation of proper
payload in PutObjectRequest and UploadPartRequest.

Fixes #580

### DIFF
--- a/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/S3ClientOptions.java
+++ b/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/S3ClientOptions.java
@@ -22,13 +22,20 @@ public class S3ClientOptions {
     /** The default setting for use of path-style access */
     public static final boolean DEFAULT_PATH_STYLE_ACCESS = false;
 
+    /** The default setting for use of chunked encoding */
+    public static final boolean DEFAULT_CHUNKED_ENCODING_DISABLED = false;
+
     /** Flag for use of path-style access */
     private boolean pathStyleAccess = DEFAULT_PATH_STYLE_ACCESS;
+
+    /** Flag for user of chunked encoding */
+    private boolean chunkedEncodingDisabled = DEFAULT_CHUNKED_ENCODING_DISABLED;
 
     public S3ClientOptions() {}
 
     public S3ClientOptions( S3ClientOptions other ) {
         this.pathStyleAccess = other.pathStyleAccess;
+        this.chunkedEncodingDisabled = other.chunkedEncodingDisabled;
     }
 
     /**
@@ -51,6 +58,28 @@ public class S3ClientOptions {
      */
     public boolean isPathStyleAccess() {
         return pathStyleAccess;
+    }
+
+    /**
+     * <p>
+     * Returns whether the client has chunked encoding disabled for all
+     * requests.
+     * </p>
+     * <p>
+     * The default behaviour is to detect API request which enables chunked
+     * encoding automatically for PutObjectRequest and UploadPartRequest.
+     * Setting this flag will result in disabling chunked encoding for all
+     * requests.
+     * Note: Enabling this option has performance implications since each
+     * payload will have to be checksum calculated. If your payload is large
+     * this will affect the overall time required to upload an object.
+     * Using this option is recommended only if your endpoint does not
+     * implement chunked uploading.
+     * </p>
+     * @return True is the client should always enable chunked encoding.
+     */
+    public boolean isChunkedEncodingDisabled() {
+        return chunkedEncodingDisabled;
     }
 
     /**
@@ -100,6 +129,29 @@ public class S3ClientOptions {
      */
     public S3ClientOptions withPathStyleAccess(boolean pathStyleAccess) {
       setPathStyleAccess(pathStyleAccess);
+      return this;
+    }
+
+    /**
+     * <p>
+     * Configures the client to disable chunked encoding for all requests.
+     * </p>
+     * <p>
+     * The default behaviour is to detect API request which enables chunked
+     * encoding automatically for PutObjectRequest and UploadPartRequest.
+     * Setting this flag will result in disabling chunked encoding for all
+     * requests.
+     * Note: Enabling this option has performance implications since each
+     * payload will have to be checksum calculated. If your payload is large
+     * this will affect the overall time required to upload an object.
+     * Using this option is recommended only if your endpoint does not
+     * implement chunked uploading.
+     * </p>
+     * @return The updated S3ClientOptions object with chunked encoding
+     *         disabled setting.
+     */
+    public S3ClientOptions disableChunkedEncoding() {
+      this.chunkedEncodingDisabled = true;
       return this;
     }
 
--- a/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/internal/AWSS3V4Signer.java
+++ b/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/internal/AWSS3V4Signer.java
@@ -36,13 +36,15 @@ import com.amazonaws.util.BinaryUtils;
  */
 public class AWSS3V4Signer extends AWS4Signer {
     private static final String CONTENT_SHA_256 = "STREAMING-AWS4-HMAC-SHA256-PAYLOAD";
+    private static boolean isChunkedEncodingDisabled;
 
     /**
      * Don't double-url-encode path elements; S3 expects path elements to be
      * encoded only once in the canonical URI.
      */
-    public AWSS3V4Signer() {
+    public AWSS3V4Signer(boolean isChunkedEncodingDisabled) {
         super(false);
+        this.isChunkedEncodingDisabled = isChunkedEncodingDisabled;
     }
 
     /**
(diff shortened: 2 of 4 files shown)


aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/internal/S3V4AuthErrorRetryStrategy.java
```diff
@@ -107,7 +107,9 @@ private AuthRetryParameters redirectToS3External() {
     }
 
     private AWSS3V4Signer buildSigV4Signer(final String region) {
-        AWSS3V4Signer v4Signer = new AWSS3V4Signer();
+        // Retry with chunkedEncoding set to true.
+        boolean isChunkedEncoding = true;
+        AWSS3V4Signer v4Signer = new AWSS3V4Signer(isChunkedEncoding);
```
shorea: For your use case this class will never be used so this is fine but it may be confusing if another developer sets the configuration option and it isn't honored in the retry policy.

aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/internal/AWSS3V4Signer.java
```diff
@@ -51,7 +53,7 @@ public AWSS3V4Signer() {
     @Override
     protected void processRequestPayload(SignableRequest<?> request, byte[] signature,
             byte[] signingKey, AWS4SignerRequestParams signerRequestParams) {
-        if (useChunkEncoding(request)) {
+        if (useChunkEncoding(request) && this.isChunkedEncoding) {
```
shorea: Can you move the second part of the conditional into useChunkEncoding?

aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/S3ClientOptions.java
```diff
@@ -55,6 +62,22 @@ public boolean isPathStyleAccess() {
...
+     * </p>
+     * <p>
+     * The default behaviour is to detect API request which enables chunked
+     * encoding automatically for PutObjectRequest and UploadPartRequest.
+     * Setting this flag will result in disabling chunked encoding for all
+     * requests.
+     * </p>
+     * @return True is the client should always enable chunked encoding.
+     */
+    public boolean isChunkedEncoding() {
```
shorea: Thinking more about this, it might make sense to expose the property just as disableChunkedEncoding. I think it may be confusing if a developer calls setUseChunkedEncoding(true) and we only don't always do chunked encoding (i.e. we only do chunked encoding for PutObjectRequest or UploadPartRequest). It's also less clear that the default is enabled.

So how about 

isChunkedEncodingDisabled()
disableChunkedEncoding()

aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/S3ClientOptions.java
```diff
@@ -103,4 +143,22 @@ public S3ClientOptions withPathStyleAccess(boolean pathStyleAccess) {
...
+     * <p>
+     * The default behaviour is to detect API request which enables chunked
+     * encoding automatically for PutObjectRequest and UploadPartRequest.
+     * Setting this flag will result in disabling chunked encoding for all
+     * requests.
+     * </p>
+     * @return The updated S3ClientOptions object with chunked encoding
+     *         disabled setting.
+     */
+    public S3ClientOptions disableChunkedEncoding() {
```
shorea: @manikandanrs suggested we clearly call out in the documentation that this has performance implications when using the real S3. Could you add a blurb saying this is not recommended for typical use cases and should only be used when the S3 implementation doesn't support chunked encoding?

aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/S3ClientOptions.java
harshavardhana: > isChunkedEncodingDisabled()
> disableChunkedEncoding()

Makes sense.

aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/S3ClientOptions.java
harshavardhana: > @manikandanrs suggested we clearly call out in the documentation that this has performance implications when using the real S3. Could you add a blurb saying this is not recommended for typical use cases and should only be used when the S3 implementation doesn't support chunked encoding?

Yes will do.

harshavardhana:
> For your use case this class will never be used so this is fine but it may be confusing if another developer sets the configuration option and it isn't honored in the retry policy.

Are you saying i don't do this? - i need to create a separate overloaded class to avoid compiler errors. if you don't wish to set the value here. 

the only reason i did that because i couldn't find a cleaner way to pass down s3ClientOptions. Should i pass this value in EndpointResolver? it doesn't look like the right place.

aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/S3ClientOptions.java
shorea: It just seems like it would be surprising if a customer has explicitly disabled chunked encoding but it was magically re-enabled in the auth error retry strategy. You should be able to pass the client options in the constructor of this class. It's created in the invoke method of AmazonS3Client.

harshavardhana: > It just seems like it would be surprising if a customer has explicitly disabled chunked encoding but it was magically re-enabled in the auth error retry strategy. You should be able to pass the client options in the constructor of this class. It's created in the invoke method of AmazonS3Client.

Yes understood, let me add another variable and take it.

harshavardhana:
Need to test this properly locally, please do not merge.

shorea:
Hold off on testing. I had to rethink how this was implemented. I'll open a PR to show the changes and we can discuss if that meets your needs.

harshavardhana:
> Hold off on testing. I had to rethink how this was implemented. I'll open a PR to show the changes and we can discuss if that meets your needs.

Oh okay cool.

shorea:
Just created a pull request. During testing discovered this would not work for our SignerFactory as it expects a default constructor to be found and there's no clean way to get the S3ClientOptions to the Factory to instantiate the signer with the right configuration. The SignerFactory would be used to create the V4 signer if the signerRegionOverride was set. I'm guessing your customers are using the ENFORCE_SIG_V4 system property to enable SigV4 for S3 so you wouldn't have hit a code path that instantiates the SigV4 signer via SignerFactory. This change includes your new option in the request so that it's available no matter how the signer is created, via the factory or explicitly via a call to it's constructor. Let me know what you think and if this meets your needs. This PR isn't intended to be merged, if it's satisfactory I'll accept your PR and stage the changes internally.

harshavardhana:
> Just created a pull request. During testing discovered this would not work for our SignerFactory as it expects a default constructor to be found and there's no clean way to get the S3ClientOptions to the Factory to instantiate the signer with the right configuration. The SignerFactory would be used to create the V4 signer if the signerRegionOverride was set. I'm guessing your customers are using the ENFORCE_SIG_V4 system property to enable SigV4 for S3 so you wouldn't have hit a code path that instantiates the SigV4 signer via SignerFactory. This change includes your new option in the request so that it's available no matter how the signer is created, via the factory or explicitly via a call to it's constructor. Let me know what you think and if this meets your needs. This PR isn't intended to be merged, if it's satisfactory I'll accept your PR and stage the changes internally.

This is perfect thank you +1 on your change.

shorea:
Awesome! I'll run the integration tests over night and get this merged in tomorrow if all goes well.

harshavardhana:
> Awesome! I'll run the integration tests over night and get this merged in tomorrow if all goes well.

Cool let me know if you need me to do something on my end.

shorea:
Can you confirm in the comments of this PR that you are submitting the code under the Apache 2.0 license?

harshavardhana:
> Can you confirm in the comments of this PR that you are submitting the code under the Apache 2.0 license?

Yes i confirm this patch is submitted under Apache 2.0 License.

shorea:
BTW tests passed, just waiting for internal approval and I'll get this merged in.

shorea:
Approved. Thanks for the contribution! This will be available in the first release of 2016 :)

harshavardhana:
> Approved. Thanks for the contribution! This will be available in the first release of 2016 :)

No thank you. :-)

PR #589, opened 2015-12-29
Copy the list of ETags before sorting

colinmarc:
Some S3 multipart upload implementations (for example, the new hadoop
S3 code) pass an immutable list of ETags to `complete`. This can
cause problems when we try to sort the ETags for inclusion in the
actual request.

We can sidestep that by just copying the list before we try to sort
it.

Thanks in advance!


manikandanrs:
Change looks good to me. Can you confirm if you are submitting under Apache 2.0 license ?

colinmarc:
Sure! Happy to submit under whatever license.

manikandanrs:
Thanks for confirming that you are submitting under Apache 2.0 license. We need this to keep legal happy. I will get this merged as part of next release.

PR #590, opened 2016-01-04
Process headers case-insensitive

hnejra:
According to RFC 2616 section 4.2 HTTP header fields should be processed case insensitive (http://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2):
"Each header field consists of a name followed by a colon (":") and the field value. Field names are case-insensitive."

This is currently not being done by the aws-java-sdk-s3 module, causing file uploads to S3 compatible storage providers such as Riak CS to fail. The reason they fail is because the response headers from Riak CS have a "Etag" field, while the aws-java-sdk-s3 module expects "ETag".

The attached changes fix this by handling all headers case insensitive. The used TreeMap will be a bit slower than a HashMap, but this should be in the range of nano-seconds, which I would not expect to be a bottleneck. If there are scenarios I cannot oversee right now, which do make this a bottleneck, all headers could first be put to lowercase and then stored in and retrieved from the HashMap, but this looks less elegant in code in my opinion.

### DIFF
--- a/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/internal/AbstractS3ResponseHandler.java
+++ b/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/internal/AbstractS3ResponseHandler.java
@@ -110,29 +110,29 @@ public abstract class AbstractS3ResponseHandler<T>
                 metadata.addUserMetadata(key, header.getValue());
             } else if (ignoredHeaders.contains(key)) {
                 // ignore...
-            } else if (key.equals(Headers.LAST_MODIFIED)) {
+            } else if (key.equalsIgnoreCase(Headers.LAST_MODIFIED)) {
                 try {
                     metadata.setHeader(key, ServiceUtils.parseRfc822Date(header.getValue()));
                 } catch (Exception pe) {
                     log.warn("Unable to parse last modified date: " + header.getValue(), pe);
                 }
-            } else if (key.equals(Headers.CONTENT_LENGTH)) {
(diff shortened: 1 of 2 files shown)


hnejra:
3/4 integration jobs succeeded, only the one for oraclejdk7 failed, because it went out of memory:
[ERROR] The system is out of resources.
[ERROR] Consult the following stack trace for details.
[ERROR] java.lang.OutOfMemoryError: PermGen space

I don't think this is related to the changes of this pull request, but I cannot retrigger the job to verify.

shorea:
Thanks for the contribution. Running this through our integration tests now. Can you confirm in the comments of this PR that you are submitting under the Apache 2.0 license?

hnejra:
@shorea: Thanks for verifying the tests. Yes, I am submitting this under the Apache 2.0 license.

shorea:
Good news, tests passed! Waiting for internal approval then I'll get this merged in.

hnejra:
Thanks! 
I just found out we need an additional change to make checking the response header keys work case insensitive as well here: https://github.com/aws/aws-sdk-java/blob/1.9.21/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/internal/AbstractS3ResponseHandler.java#L125
We need to use equalsIgnoreCase() here. I'll make that change and add that to this pull request.

hnejra:
I made the change for checking the response header keys, this can be reviewed as well now.

shorea:
Tests passed with new changes and received internal approval. Merging in. Thanks for the contribution!

hnejra:
Great! I'm glad I could contribute.

PR #594, opened 2016-01-06
Fix typo in SNS SignatureChecker causing verification of SubscriptionConfirmation messages to fail

tpunder:
The SNS SignatureChecker class doesn't currently work with `SubscriptionConfirmation` messages because the `SubscribeURL` is not included as part of the signature verification due to a typo in the code.

The `INTERESTING_FIELDS` set currently includes `SUBSCRIBE_TYPE` which is defined as:

    private final String SUBSCRIBE_TYPE = "SubscriptionConfirmation";

This doesn't make any sense because `SubscriptionConfirmation` is not a field name.  It's a value of the `Type` field.

Instead of `SUBSCRIBE_TYPE` it should be `SUBSCRIBE_URL`.


shorea:
Change looks good, thanks for the contributon! Can you confirm in the comments of the PR that this is being submitted under the Apache 2.0 license.

tpunder:
Yes, I am submitting this under the Apache 2.0 license.

shorea:
Thanks for the contribution! This will be available in our next release.

PR #595, opened 2016-01-07
Make chunked encoding option non-static

andrewgaul:
Previously disabling chunked encoding set a static which affected all
instances of `AWSS3V4Signer`, not just the one configured by
`AmazonS3Client.setS3ClientOptions`.  References #580.


andrewgaul:
@harshavardhana Please review.

shorea:
Changes look good, thanks for the fix! Can you confirm you are submitting under Apache 2.0?

andrewgaul:
Confirmed that I submit this change under the Apache 2.0 license.

harshavardhana:
LGTM  @andrewgaul

shorea:
Actually on closer inspection this should be fixed per the latest release. The disableChunkedEncoding is no longer a field in the signer but passed in the request context. @andrewgaul could you retest with 1.10.45.

andrewgaul:
Confirmed fixed in 1.10.45 without this pull request.  Part of the squashed commit 052cdb984b57f90615b9c40f9a743cefaa76cfcd includes similar functionality.

Issue #524, opened 2015-09-30
SDK not honoring https.proxyHost and https.proxyPort from Java


jfisbein:
Even when launching applications that use aws-sdk with 
`-Dhttps.proxyHost=1.2.3.4 -Dhttps.proxyPort=3128` [command line options](http://docs.oracle.com/javase/8/docs/api/java/net/doc-files/net-properties.html), all https calls are not using the proxy as they should.

manikandanrs:
Currently SDK supports configuring proxy settings via [ClientConfiguration#setProxyPort] (http://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/ClientConfiguration.html#setProxyPort(int)) and [ClientConfiguration#setProxyHost] (http://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/ClientConfiguration.html#setProxyHost(java.lang.String)).

We will discuss this internally and let you know if this can be considered as a feature request.

manikandanrs:
An update here. The team agreed that it is nice to have feature on the Java SDK. I am turning this issue into feature request. 

The SDK will always take into consideration the proxy settings in ClientConfiguration and will only check the https.proxyHost and https.proxyPort only if the proxy settings are not set in ClientConfiguration. 

Note: We accepts pull requests from our customers. If you wish to contribute one for this feature, feel free to do so and we would be happy to take a look at it and merge.

bobby:
@manikandanrs I've added a PR which addresses this issue, and I'd love your thoughts on how to incorporated it into a future release.

manikandanrs:
@bobby 

Thanks for the contribution. I will take a look at your PR.

manikandanrs:
This is fixed as part of #596

SushmitaGoswami:
Is there any alternatives to ClientConfiguration#setProxyPort and ClientConfiguration#setProxyHost in aws java sdk 2.x? I was able to do this in aws sdk 1.x, but found no way to do this in aws java sdk 2.x

debora-ito:
@SushmitaGoswami In SDK 2.x, proxy settings are defined using the [ProxyConfiguration](https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/http/apache/ProxyConfiguration.html) in the `HttpClient` configuration, so it would be something like:

```java
SdkHttpClient httpClient = ApacheHttpClient.builder()
        .proxyConfiguration(
[...]

PR #596, opened 2016-01-08
ClientConfiguration checks Java proxy properties

bobby:
Addresses #524:

* Always prefers set configuration variables over java system properties, but if unset:
  * Uses values of `https.proxy(Host|Port|User|Password)` if protocol is `Protocol.HTTPS`
  * Uses values of `http.proxy(Host|Port|User|Password)` if protocol is `Protocol.HTTP`

`http(s).proxyHost` and `http(s).proxyPort` are official parts of the [Java API](https://docs.oracle.com/javase/7/docs/api/java/net/doc-files/net-properties.html), whereas `http(s).proxyUser` and `http(s).proxyPassword` are traditions in the Java community.

Two useful follow-ons to this pull request that would really put a bow on #524 would be:

1. Passing `http.nonProxyHosts` (no `https` variant, as per spec) through to the clients somehow (useful e.g. to bypass proxy when hitting 169.254.169.254 in an AWS context, but widely used in general).  This addition could possibly be higher impact, as it introduces a new configuration that must be passed into the underlying HTTP client.
2. Checking, parsing, and using the widely used environment variables `HTTPS_PROXY`, `https_proxy`, `HTTP_PROXY`, `http_proxy`, `NO_PROXY`, and `no_proxy` (where the last two serve the same purpose as `http.nonProxyHosts`).  Precedence order would be: configured values, then Java system properties, then environment variables.  This change (modulo `NO_PROXY`) would be rather straightforward.

Are the above recommendations of interest?  I'd be happy to provide further pull requests.

### DIFF
--- a/aws-java-sdk-core/src/main/java/com/amazonaws/ClientConfiguration.java
+++ b/aws-java-sdk-core/src/main/java/com/amazonaws/ClientConfiguration.java
@@ -466,11 +466,7 @@ public class ClientConfiguration {
      * @return The proxy host the client will connect through.
      */
     public String getProxyHost() {
-        if (proxyHost != null) {
-            return proxyHost;
-        } else {
-            return getProxyHostProperty();
-        }
+        return (proxyHost != null) ? proxyHost : getProxyHostProperty();
     }
 
     /**
@@ -500,15 +496,11 @@ public class ClientConfiguration {
      * Returns the Java system property for proxy port depending on
      * {@link this.getProtocol()}: i.e. if protocol is https, returns
      * the value of the system property https.proxyPort, otherwise
-     * returns value of http.proxyPort.
+     * returns value of http.proxyPort.  Defaults to {@link this.proxyPort}
+     * if the system property is not set with a valid port number.
      */
     private int getProxyPortProperty() {
-        String proxyPortString;
-        if (getProtocol() == Protocol.HTTPS) {
-            proxyPortString = System.getProperty("https.proxyPort");
-        } else {
-            proxyPortString = System.getProperty("http.proxyPort");
-        }
+        String proxyPortString = (getProtocol() == Protocol.HTTPS) ? System.getProperty("https.proxyPort") : System.getProperty("http.proxyPort");
         try {
             return Integer.parseInt(proxyPortString);
         } catch (NumberFormatException e) {
@@ -527,11 +519,7 @@ public class ClientConfiguration {
      * @return The proxy port the client will connect through.
      */
     public int getProxyPort() {
-        if (proxyPort > 0) {
-            return proxyPort;
-        } else {
-            return getProxyPortProperty();
-        }
+        return (proxyPort >= 0) ? proxyPort : getProxyPortProperty();
     }
 
     /**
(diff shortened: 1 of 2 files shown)


jdamick:
+1 It would nice if it supported env HTTP/S_PROXY settings too

aws-java-sdk-core/src/main/java/com/amazonaws/ClientConfiguration.java
```diff
@@ -442,12 +442,35 @@ public ClientConfiguration withLocalAddress(InetAddress localAddress) {
...
+     * if not provided, checks the value of the Java system property
+     * for proxy host according to {@link this.getProtocol()}: i.e. if
+     * protocol is https, returns the value of the system property
+     * https.proxyHost, otherwise returns value of http.proxyHost.
      *
      * @return The proxy host the client will connect through.
      */
     public String getProxyHost() {
-        return proxyHost;
+        if (proxyHost != null) {
```
manikandanrs: Minor: Can you use the ternary operator here.

aws-java-sdk-core/src/main/java/com/amazonaws/ClientConfiguration.java
```diff
@@ -474,12 +497,41 @@ public ClientConfiguration withProxyHost(String proxyHost) {
     }
 
     /**
-     * Returns the optional proxy port the client will connect through.
+     * Returns the Java system property for proxy port depending on
+     * {@link this.getProtocol()}: i.e. if protocol is https, returns
+     * the value of the system property https.proxyPort, otherwise
+     * returns value of http.proxyPort.
```
manikandanrs: Would be nice if we can have the javadoc updated with "Defaults to proxyPort is the System property is not set with a valid port number"

aws-java-sdk-core/src/main/java/com/amazonaws/ClientConfiguration.java
```diff
@@ -474,12 +497,41 @@ public ClientConfiguration withProxyHost(String proxyHost) {
...
+     * if not provided, checks the value of the Java system property
+     * for proxy port according to {@link this.getProtocol()}: i.e. if
+     * protocol is https, returns the value of the system property
+     * https.proxyPort, otherwise returns value of http.proxyPort.
      *
      * @return The proxy port the client will connect through.
      */
     public int getProxyPort() {
-        return proxyPort;
+        if (proxyPort > 0) {
```
manikandanrs: proxyPort >= 0 ?

manikandanrs:
PR looks great .Can you add unit tests to the PR ?

bobby:
@manikandanrs: I've incorporated your recommendations in my last commit.

manikandanrs:
Thanks for pull request. Can you confirm if you are submitting under Apache 2.0 license ?

bobby:
I confirm that I submit this pull request under the terms of the Apache 2.0 license.

manikandanrs:
Thanks for the contribution. This has been staged for the next release.

bobby:
Thanks, looking forward to it!  Thanks for the quick turn around on this PR, kudos to you and your team, @manikandanrs!

benkropf:
@bobby just curious, did you end up implementing 2 from your suggested follow-ons? My app currently relies on those system environment variables which I dont believe the SDK currently honors, unless its been changed on a newer version

PR #598, opened 2016-01-08
S3 system property to disable sigv4 upgrade

sjones4:
This pull request adds a system property that can be used to disable upgrade to signature version 4 for S3.

The use case for this is against S3 clones that do not support signature v4 where code modifications are not possible (the signer override property of the client configuration can be used otherwise)

The system property does not disable signature version 4, only upgrades to signature v4, so in addition to the system property an override configuration may be required (awssdk_config_override.json)

### DIFF
--- a/aws-java-sdk-core/src/main/java/com/amazonaws/SDKGlobalConfiguration.java
+++ b/aws-java-sdk-core/src/main/java/com/amazonaws/SDKGlobalConfiguration.java
@@ -115,6 +115,16 @@ public class SDKGlobalConfiguration {
         "com.amazonaws.services.s3.enforceV4";
 
     /**
+     * Like {@link #ENFORCE_S3_SIGV4_SYSTEM_PROPERTY}, but causes the client to
+     * never upgrade to Signature Version 4. This property is overridden by
+     * {@link #ENFORCE_S3_SIGV4_SYSTEM_PROPERTY} and Signature Version 4 will
+     * still be used when otherwise configured or overridden for a client
+     * instance.
+     */
+    public static final String DISABLE_S3_SIGV4_UPGRADE_SYSTEM_PROPERTY =
+        "com.amazonaws.services.s3.disableV4Upgrade";
+
+    /**
      * @deprecated with {@link AmazonWebServiceRequest#getRequestClientOptions()}
      * and {@link RequestClientOptions#setReadLimit(int)}.
      * <p>
--- a/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/AmazonS3Client.java
+++ b/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/AmazonS3Client.java
@@ -14,6 +14,7 @@
  */
 package com.amazonaws.services.s3;
 
+import static com.amazonaws.SDKGlobalConfiguration.DISABLE_S3_SIGV4_UPGRADE_SYSTEM_PROPERTY;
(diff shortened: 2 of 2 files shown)


shorea:
Can you provide an example where the signer override isn't feasible to use?

sjones4:
A couple of examples come to mind. In the first case you would have an application that runs against AWS/S3 and some other implementation of the S3 API (such as a Eucalyptus cloud) and would want to run the same code against both, changing only the configuration/deployment. Another case would be where an organization runs (AWS/S3 compatible) software provided by a third-party (that uses the AWS SDK) and wants to be able to configure that software to run against another S3 implementation, in this case they would not have access to the source code for the application.

shorea:
Wouldn't you need access to the client to override the endpoint to point towards the other implementation?

sjones4:
Yes, you would also need to configure the endpoint. In the first example, you would ensure you wrote the application so the endpoint could be configured, in the second example you would require the 3rd party software to make the endpoint configurable.

shorea:
I'm just thinking that configuring the client with code is more flexible then a system property. Properties are JVM wide so you can't talk to the real S3 and a different impl in the same application with this approach. If you already have to configure the endpoint I think it would be more symmetric if you had to configure the signer through code rather than a system property.

sjones4:
Typically the endpoint and credentials are things the user can already configure. There is no way to configure the signer without code change due to the special override handling in the client that means the configured values ("awssdk_config_override.json") are ignored.

shorea:
Didn't quite understand the last sentence. Do you mean config specified in the override file is not honored? If so this sounds like a bug.

I guess what I'm getting at is that if we already provide a way to enforce a specific signer through the signer override and you have access to the client (either direct access in the first case where you are creating it yourself or some kind of proxied access when using a library that creates it internally but allows you to configure it) then you can already write code that can handle both implementations without a code change. I.E. something like this

```java
AmazonS3Client s3;
// The value of useFakeS3 can come from some external configuration source
if(useFakeS3) {
   s3 = new AmazonS3Client(new ClientConfiguration().withSignerOverride("foo");
} else {
   s3 = new AmazonS3Client();
}
```

sjones4:
I agree that if you are writing code it is possible to configure the signature version for any action/region. 

What would be useful is to be able to set the signer using the regular configuration mechanism ("awssdk_config_override.json") and not have that configuration overridden unconditionally by the AmazonS3Client for some actions/regions as currently occurs in AmazonS3Client for upgrade to signature version 4 (AmazonS3Client#createSigner)

shorea:
Sorry for the delay in response @sjones4 . Was chatting with a colleague about this and forgot to respond on the PR. We've actually got some changes in the works that would make the upgrade path redundant and the client will just honor what's in the config. This work is in our current sprint so I'm hesitant to add the property as it will be unnecessary soon. What's your timeline for this fix? Do you have a suitable workaround until we can get this rolled out?

sjones4:
The client using whatever is configured works for me and is preferable to requiring another system property.

How will I know when those changes are rolled out so I can test my use case?

shorea:
I'll make a note on the task to update this PR when it's rolled out.

sjones4:
Sounds good. Thanks for your help with this issue.

shorea:
I apologize for the long silence, this got deprioritized. We are looking to fix this in an upcoming major version bump and remove the properties altogether.

shorea:
This is no longer an issue in version 1.11 of the SDK.

Issue #602, opened 2016-01-15
LambdaInvokerFactory and @LambdaFunction do not support multiple environments


codan84:
If I create an interface:
```
interface MyLambdas {
    @LambdaFunction(functionName = "my-aws-lambda-name")
    public String helloWorld(String name);
}
```

And then build an implementation:
```
AWSLambdaClient lambda = new AWSLambdaClient(); lambda.configureRegion(Regions.EU_WEST_1);
MyLambdas lambdas = LambdaInvokerFactory.build(MyLambdas.class, lambda);
```

This is all fine. However, it doesn't support different deployment environments (say int/test/live), where a lambda can have different name (say created via CloudFormation). 

To address this, I would propose to overload `LambdaInvokerFactory.build` to take an additional map of key->lambdaName. Key would store the value passed to `@LambdaFunction`, whilst value for this key would be the actual function name in AWS. This would allow to use different function names for different environments. An example of use:
```
interface MyLambdas {
    @LambdaFunction(functionName = "hello-world-lambda-key")
    public String helloWorld(String name);
}

Map<String, String> myLambdaNames = new HashMap<String, String>();
myLambdaNames.put("hello-world-lambda-key", "my-aws-lambda-name");

AWSLambdaClient lambda = new AWSLambdaClient();
lambda.configureRegion(Regions.EU_WEST_1);
MyLambdas lambdas = LambdaInvokerFactory.build(MyLambdas.class, lambda, myLambdaNames);
```

As you can see this let's us to pull the names at runtime, from places like config etc. I think this would be a beneficial addition to the API.  
I have implemented this functionality in [this fork](https://github.com/codan84/aws-sdk-java)

shorea:
Made some comments in the PR.

gigi888:
so cool! it solves our problem too! Many thanks

PR #603, opened 2016-01-15
LambdaInvokeFactory supporting lambda names from a map

codan84:
The issue this PR aims to address is [described here](https://github.com/aws/aws-sdk-java/issues/602)

### DIFF
--- a/aws-java-sdk-lambda/src/main/java/com/amazonaws/services/lambda/invoke/LambdaFunctionNameResolver.java
+++ b/aws-java-sdk-lambda/src/main/java/com/amazonaws/services/lambda/invoke/LambdaFunctionNameResolver.java
@@ -1,0 +1,7 @@
+package com.amazonaws.services.lambda.invoke;
+
+import java.lang.reflect.Method;
+
+public interface LambdaFunctionNameResolver {
+    String getFunctionName(Method method, LambdaFunction annotation, LambdaInvokerFactoryConfig config);
+}
--- a/aws-java-sdk-lambda/src/main/java/com/amazonaws/services/lambda/invoke/LambdaInvokerFactory.java
+++ b/aws-java-sdk-lambda/src/main/java/com/amazonaws/services/lambda/invoke/LambdaInvokerFactory.java
@@ -14,18 +14,6 @@
  */
 package com.amazonaws.services.lambda.invoke;
 
-import java.io.IOException;
-import java.lang.reflect.Constructor;
-import java.lang.reflect.InvocationHandler;
-import java.lang.reflect.Method;
-import java.lang.reflect.Proxy;
-import java.lang.reflect.Type;
-import java.nio.ByteBuffer;
-import java.util.List;
-
-import org.apache.commons.logging.Log;
-import org.apache.commons.logging.LogFactory;
-
 import com.amazonaws.services.lambda.AWSLambda;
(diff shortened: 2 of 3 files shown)


shorea:
I think it would be simpler and more flexible just to take an override for the name rather then a map. The caller can then choose between pulling the name from a map, loading it from some external configuration source, or computing it dynamically from some context. Also we can probably avoid a subclass of the invocation handler and just pass the derived function name (whether that comes from the annotation or the override) into the handler class. It seems very possible that we will add more overrides and configuration options in the future so I think it would be prudent to wrap the function name override in some kind of configuration object (LambdaInvokerFactoryConfig for example) so that we can easily add fields to it without having an explosion of build method overloads, Similiar to what we do with the DynamoDBMapper (http://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/dynamodbv2/datamodeling/DynamoDBMapper.html#DynamoDBMapper%28com.amazonaws.services.dynamodbv2.AmazonDynamoDB,%20com.amazonaws.services.dynamodbv2.datamodeling.DynamoDBMapperConfig%29).

codan84:
Hi Andrew, I have looked at [DynamoDBMapperConfig.TableNameOverride](http://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/dynamodbv2/datamodeling/DynamoDBMapperConfig.TableNameOverride.html) and indeed this seems like a better approach. I will work on a similar solution.

shorea:
Yeah I actually wasn't thinking about this in particular but I like your idea better.

Few comments inline
LambdaNameResolver -> LambdaFunctionNameResolver

Rename method and add some additional context params. The default resolver would just use the class and grab the annotation. A custom one might do something fancier.
getName(String key) -> getFunctionName(Class<?> clzz, LambdaInvokerFactoryConfig config)

Could you also confirm you are submitting this under the Apache 2.0 license?

codan84:
Sounds like a plan. I will try to work on this today. And yes, I will do it under Apache 2.0.

codan84:
I think the work is done, unless you have some further comments :)

shorea:
Change looks good, we've got a release going now but I'll get this merged in and staged for the next release.

shorea:
This has been merged. Thanks for the contribution! This will be available in the next release of the SDK.

shorea:
@codan84, we have a release scheduled for later today but unfortunately this change won't be included as I screwed up and didn't stage it correctly. The diff for today will look like this is being removed but that's just an artifact of how we publish changes to Github. We will have another release of the SDK next week that will add these changes back in and make them publicly available. I apologize for the inconvenience. We are excited for these changes though, I think customers will find a lot of value in it.

codan84:
@shorea Will this make it to 1.10.51? And do you know when is the release planned? We could really do with this functionality... Cheers.

shorea:
Yep we've got a release tentatively scheduled for today.

strokine:
LambdaInvokerFactoryConfig is depricated

shorea:
Check out the new builder.

https://github.com/aws/aws-sdk-java/blob/master/aws-java-sdk-lambda/src/main/java/com/amazonaws/services/lambda/invoke/LambdaInvokerFactory.java#L110

AzeemBOL:
Great... Worked for me.

meeza:
great.. worked for me.

PR #620, opened 2016-01-28
use bean pattern for disabling chunked encoding

notnoopci:
Current implementation of disabling chunk encoding violates the bean pattern of `isFoo` and `withFoo` that's pretty ubitiques in the SDK and is currently relied on by some common libraries, e.g. [Amazonica](https://github.com/mcohen01/amazonica).

This deprecates `disableChunkedEncoding` function and replaces it with `setChunkedEncodingDisabled` and `withChunkedEncodingDisabled` methods to match the path style access functions.

### DIFF
--- a/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/S3ClientOptions.java
+++ b/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/S3ClientOptions.java
@@ -107,6 +107,28 @@ public class S3ClientOptions {
 
     /**
      * <p>
+     * Configures the client to disable chunked encoding for all requests.
+     * </p>
+     * <p>
+     * The default behavior is to enable chunked encoding automatically for PutObjectRequest and
+     * UploadPartRequest. Setting this flag will result in disabling chunked encoding for all
+     * requests.
+     * </p>
+     * <p>
+     * <b>Note:</b> Enabling this option has performance implications since the checksum for the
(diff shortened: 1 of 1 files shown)


shorea:
Can you explain why Amazonica relies on the bean convention? Just want to know for my own benefit so that we are more careful about this in the future.

notnoopci:
Amazonica converts the function and class names to be idiomatic clojure naming convention and patterns of using simple maps for representing aws specific data.  The conversion process heavily leans on reflection and bean convention.  For example, if the following map is used in a context that expects a `DescribeImageRequest`:

```clojure
{:owners ["self"] :image-ids ["ami-f00f9699" "ami-e0d30c89"]})
```

With reflection it will be coerced with the following equivalent Java code:

```java
DescribeImagesRequest request = new DescribeImagesRequest();
request.setOwners(Arrays.asList("self"));
request.setImageIds(Arrays.asList("ami-f00f9699", "ami-e0d30c89"));
```

You can see more details at https://github.com/mcohen01/amazonica#documentation .

The library does offer limited custom data type coercion mechanism for the exceptions - but I felt that it would be nicer to keep the AWS SDK consistent across the board over special casing it in Amazonica.

shorea:
Gotcha, thanks for clarifying. Can you confirm you are submitting under Apache 2.0?

notnoopci:
I confirm that my submission is under Apache 2.0 license.

shorea:
This has been merged. Thanks for the contribution! This will be available in the next release of the SDK.

PR #624, opened 2016-02-02
Fix issue #623 - Exception in Parallel Scan causes application to hang

### LINKED ISSUE #623: DynamoDB - Exception in Parallel Scan causes application to hang for 60 seconds
**Description:**
When call to DynamoDBMapper.parallelScan() fails with an Exception e.g. ProvisionedThroughputExceededException, the application hangs until the cached threads expire after 60 seconds.

**Example Stack trace:**
`com.amazonaws.services.dynamodbv2.model.ProvisionedThroughputExceededException: The level of configured provisioned throughput for the table was exceeded. Consider increasing your provisioning level with the UpdateTable API (Service: AmazonDynamoDBv2; Status Code: 400; Error Code: ProvisionedThroughputExceededException; Request ID: XXXXXX)
	at com.amazonaws.http.AmazonHttpClient.handleErrorResponse(AmazonHttpClient.java:1181)
	at com.amazonaws.http.AmazonHttpClient.executeOneRequest(AmazonHttpClient.java:766)
	at com.amazonaws.http.AmazonHttpClient.executeHelper(AmazonHttpClient.java:485)
	at com.amazonaws.http.AmazonHttpClient.execute(AmazonHttpClient.java:306)
	at com.amazonaws.services.dynamodbv2.AmazonDynamoDBClient.invoke(AmazonDynamoDBClient.java:1799)
	at com.amazonaws.services.dynamodbv2.AmazonDynamoDBClient.scan(AmazonDynamoDBClient.java:1519)
	at com.amazonaws.services.dynamodbv2.datamodeling.ParallelScanTask.scanNextPageOfSegment(ParallelScanTask.java:223)
	at com.amazonaws.services.dynamodbv2.datamodeling.ParallelScanTask.access$000(ParallelScanTask.java:32)
	at com.amazonaws.services.dynamodbv2.datamodeling.ParallelScanTask$1.call(ParallelScanTask.java:157)
	at com.amazonaws.services.dynamodbv2.datamodeling.ParallelScanTask$1.call(ParallelScanTask.java:149)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)`

**Expected:**
The cached thread pool used by `com.amazonaws.services.dynamodbv2.datamodeling.ParallelScanTask` should have shutdown after the failure.

paragvk:
I want to fix #623 - Exception in Parallel Scan causes application to hang

### DIFF
--- a/aws-java-sdk-dynamodb/src/main/java/com/amazonaws/services/dynamodbv2/datamodeling/ParallelScanTask.java
+++ b/aws-java-sdk-dynamodb/src/main/java/com/amazonaws/services/dynamodbv2/datamodeling/ParallelScanTask.java
@@ -163,6 +163,7 @@ public class ParallelScanTask {
                             synchronized(segmentScanStates) {
                                 segmentScanStates.set(currentSegment, SegmentScanState.Failed);
                                 segmentScanStates.notifyAll();
+                                executorService.shutdown();
                             }
                             throw e;
                         }


paragvk:
Build it again please.

shorea:
@paragvk I'm not sure we should shut down the executor if a single segment page fails. We throw the Provisioned Throughput Exception out and the user can catch, backoff a bit, and call next on the iterator and expect the mapper to try and fetch the next page of results.

paragvk:
@shorea Following are some thoughts on - could there be a compelling use case to continue after error here: 

Although, throttling errors like ProvisionedThroughputException are considered retry-able, it is hard to continue at this level of API from point of failure since `LastEvaluatedKey` is not available. Use cases like - wanting to continue with partial results OR try-catching the exception and going back to the `PaginatedParallelScanList` to get() remaining items (hoping framework to continue paging results) seem like too much responsibility for framework.

For a use case like scanning for a count, partial results are not desired and the program would want to quit. Moreover, the behaviour is same for any other non-retry-able exceptions e.g. `com.amazonaws.AmazonServiceException: The security token included in the request is expired...` where it would not make sense for caller to continue.

My use case is to scan and get exact set of eligible records (fault intolerant). In case of error (throttling or any other kind), it is desired that the program quits (but shutdown is delayed by 60 sec due to this issue).

**Short version:** Since this is the higher level API, only clean results are desired due to lack of tools like `LastEvaluatedKey`. In case of error, the framework should stop the scan and throw error. For finer control on handling errors (retry-able or not), users may be encouraged to use low level API.

PK

shorea:
Would adding an explicit shutdown on finalize satisfy your use case? If the PaginatedScanList goes out of scope and is garbage collected the finalize method will eventually be invoked and the application can shut down cleanly.

paragvk:
I am not sure, how will that be achieved. Do we have any hooks for finalize? If you could point out the place in code where this could be done, I shall try it out and let you know.

paragvk:
I tried overriding the finalize() in `ParallelScanTask` but that did not solve the issue.
```
    @Override
    protected void finalize() throws Throwable {
        super.finalize();
        executorService.shutdown();
    }
```
Alternative quick solution could be making instance `executorService` on `ParallelScanTask` accesible via getter, so that users may control it.

shorea:
Actually scratch that. ThreadPoolExecutor already does this so it won't solve the problem.

http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/8u40-b25/java/util/concurrent/ThreadPoolExecutor.java#1476

shorea:
I wonder if we could mark the threads as daemon threads and allow the application to shutdown regardless if they are still running.

shorea:
Whatcha think?

paragvk:
We could, but it wont be useful to the client requesting the scan (potentially harmful).

paragvk:
To your original [comment](https://github.com/aws/aws-sdk-java/pull/624#issuecomment-178872585) I tried something to keep going even after failure, but could not get it working. Using `PaginationLoadingStrategy.ITERATION_ONLY` configuration and a custom iterator to keep going:

 
```
 List<MyItem> scanResult = mapper.parallelScan(MyItem.class, scanExpression, numberOfSegments, mapperConfig);
        
        Iterator<MyItem> orignalIterator = scanResult.iterator();

        // Create fault tollerant iterator
        Iterator<MyItem> myFaultTollerantIterator = new Iterator<ParallelDeleteWithMapper.MyItem>() {

            @Override
            public boolean hasNext() {
                boolean hasNextElement = false;
                while (true) {
                    try {
                        hasNextElement = orignalIterator.hasNext();
                        break;
                    } catch (Exception ex) {
                        if (ex instanceof AmazonServiceException && RetryUtils.isThrottlingException((AmazonServiceException) ex)) {
                            // pause exponentially
                        } else {
                            // rethrow ex
                        }
                    }
                }
                return hasNextElement;
            }

            @Override
            public MyItem next() {
                return orignalIterator.next();
            }
        };

        // Start iterating
        while (myFaultTollerantIterator.hasNext()) {
            MyItem i = it.next();
        }
```

But i kept getting same throughput exception even if I increased the RCU to a high value while the program was running.

paragvk:
So, I am still in favour of the idea of stopping the scan (along with executor) in case of exception so the user will get the same feeling he would get while using `batchLoad()`. User who want to need more control can always use low level API.

shorea:
Sorry for the delay @paragvk . We are going through old issues and saw this is still unanswered. I'll re-investigate to see if there is a viable solution.

paragvk:
Hi @shorea this is a corner case and does not affect my normal use cases. Appreciate you checking back on this :)

shorea:
Okay so I dug into this and I agree with your assessment. When a segment fails the iterator will continue to throw the exception of that segment and the iterator becomes unusable. In that case shutting down the executor seems like the right thing to do. I'd like to write a unit test to simulate the shutdown but in the meantime could you refresh your PR with the latest SDK or create a new PR with the shutdown?

shorea:
"could you refresh your PR"

Disregard this. I see that the PR is still valid as this class hasn't changed since the PR was opened.

shorea:
This made it just in time for today's release. Merging PR

shorea:
Thanks for the contribution!

paragvk:
Thank you!

PR #634, opened 2016-02-15
Fix Topics util overwriting sqs policy

chidoli:
In Topics.subscribeQueue, adding a statement onto existing policy rather than overwriting it.


breedloj:
Hey Chi, thanks for the pull request!  It seems reasonable that some users might want this behavior.  However my concern is that this would effectively be a breaking change to those already using this method since they expect the current behavior (which is explicitly documented in the Javadoc).  What do you think about creating a new method which has this behavior or one which optionally takes in a boolean specifying whether the policy should be re-used or overwritten?  With the latter approach, we can probably mark the existing method as deprecated and eventually remove it in a future version.

chidoli:
Oh I missed that part. I read "modify policy" and thought it would extend the existing policy. I will add a method with one more parameter "boolean extendPolicy".

breedloj:
Nice!  Can you confirm that you are submitting this code under the Apache 2.0 license?

chidoli:
Yes I am aware of it.

breedloj:
This has been staged for release today and should be out shortly.  Thanks again!

PR #641, opened 2016-02-24
Implement role assumption in profile configuration

dump247:
Supports cross-account access via IAM roles.

This functionality matches the implementation in botocore (used by boto3 and the CLI). I had to copy some functionality from the STS module. I am not really happy with `ProfileRoleCredentialsProvider.Factory`, but I couldn't think of a better way to support unit testing.

### DIFF
--- a/aws-java-sdk-core/src/main/java/com/amazonaws/auth/profile/internal/ProfilesConfigFileLoader.java
+++ b/aws-java-sdk-core/src/main/java/com/amazonaws/auth/profile/internal/ProfilesConfigFileLoader.java
@@ -28,14 +28,17 @@ import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 
 import com.amazonaws.AmazonClientException;
+import com.amazonaws.auth.AWSCredentials;
 import com.amazonaws.auth.BasicAWSCredentials;
 import com.amazonaws.auth.BasicSessionCredentials;
+import com.amazonaws.auth.profile.internal.securitytoken.ProfileCredentialsService;
+import com.amazonaws.auth.profile.internal.securitytoken.RoleInfo;
 
 public class ProfilesConfigFileLoader {
 
     private static final Log LOG = LogFactory.getLog(ProfilesConfigFileLoader.class);
 
-    public static Map<String, Profile> loadProfiles(File file) {
+    public static Map<String, Profile> loadProfiles(File file, ProfileCredentialsService profileCredentialsService) {
         if (file == null) {
             throw new IllegalArgumentException(
                     "Unable to load AWS profiles: specified file is null.");
@@ -50,7 +53,7 @@ public class ProfilesConfigFileLoader {
         FileInputStream fis = null;
         try {
             fis = new FileInputStream(file);
-            return loadProfiles(fis);
+            return loadProfiles(fis, profileCredentialsService);
         } catch (IOException ioe) {
             throw new AmazonClientException(
                     "Unable to load AWS credential profiles file at: " + file.getAbsolutePath(), ioe);
@@ -65,7 +68,7 @@ public class ProfilesConfigFileLoader {
      * @param is input stream from where the profile details are read.
      * @throws IOException
      */
-    private static Map<String, Profile> loadProfiles(InputStream is) throws IOException {
+    private static Map<String, Profile> loadProfiles(InputStream is, ProfileCredentialsService profileCredentialsService) throws IOException {
         ProfilesConfigFileLoaderHelper helper = new ProfilesConfigFileLoaderHelper();
         Map<String, Map<String, String>> allProfileProperties = helper.parseProfileProperties(new Scanner(is));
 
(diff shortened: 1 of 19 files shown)


dump247:
The openjdk6 build fails because the code-generator module is set to java 1.7. Is that expected?

shorea:
Thanks for opening a pull request, this looks really interesting. I have some concerns about copying a slimmed down version of the STS client into core, I fear it will be a maintenance burden. I am also not crazy about declaring a dependency on STS in Core as forcing that on customers negates some of the benefit of a modularized SDK. What do you think about moving this into the STS client much like the existing AssumeRole and temporary credentials providers. The downside being we would never be able to add this to the default chain and you'd always have to explicitly configure the client with this provider but I think it will lead towards a cleaner and more maintainable solution.

Let us know what you think! :)

dump247:
I understand your concern, but it would be also be a significant burden if this is not part of the default provider chain. For back story, our company is moving to a role-based access model, where your default IAM permissions are limited to assuming a role. We will have separate AWS accounts for different environments and functions. That means that every application/tool/utility that uses the java sdk would need to be updated so the default chain + profile role assumption is added to every client object created. It's possible, but it's ugly.

I don't see how you can make this part of the default chain without implementing some simple sts service client. Is there anything you can think of that would help alleviate your concerns about maintenance burden? Integration tests or other ways to call out potential issues? The sts api is pretty straight forward.

dump247:
Also, we are planning on submitting change requests to other sdks, such as javascript, ruby, and go.

shorea:
So the consensus within the team is to implement this without a hard dependency on STS. The proposed approach would have a minimal STS interface in core and use reflection to get an implementation of that interface from the STS client if present. 

Some implementation guidelines
1. We should only attempt to load this interface when we encounter a profile that is using assume role and cache it for future use.
2. If we encounter a role based profile and we can't find an appropriate STS implementation then we should fail fast instead of skipping to prevent non deterministic loading of credentials.

dump247:
Just so I am clear on what you want.

Create an interface like this in core:

```java
package com.amazonaws.auth.profile.internal.securitytoken;

public interface SimpleTokenCredentialsService {
  Credentials assumeRole(RoleInfo roleInfo);
}
```

Implement something that attempts to load class `com.amazonaws.services.securitytoken.AWSSecurityTokenServiceClient` and provide an adapter implementation of `SimpleTokenCredentialsService` on top of it.

If role assumption is configured for any profile and `com.amazonaws.services.securitytoken.AWSSecurityTokenServiceClient` is not on the class path (the loader mentioned above doesn't find the class), raise an exception.

Basically, if the sts module is not on the class path and role assumption is configured in a profile, even if you don't use the profile, you get an error.

Let me know if the type naming and structure seems ok or if you have some recommendations for that.

dump247:
I rewrote the profile assume role functionality to something that I think matches your requirements. One thing I don't follow is that the sts dependency on the core is the previous release version and not the snapshot. Let me know what you think and I will get onto writing tests as well.

aws-java-sdk-sts/pom.xml
```diff
@@ -16,7 +16,7 @@
     <dependency>
       <groupId>com.amazonaws</groupId>
       <artifactId>aws-java-sdk-core</artifactId>
-      <version>1.10.57</version>
+      <version>1.10.58-SNAPSHOT</version>
```
dump247: I am not sure what to do here.

shorea: Yeah this is annoying, I wouldn't worry about it too much. We will actually overwrite this when we release so do whatever allows you to test this.

shorea:
Your assessment sounds good to me. Is this ready for review?

dump247:
Implementation is ready for review. I still need to add tests since I wanted to make sure that the structure and implementation were acceptable first.

shorea:
What do you think about exposing a factory for STSAssumeRoleSessionCredentialsProvider as the interface rather then the service client?

dump247:
Sounds like a good idea. I'm going to be working on that today.

dump247:
@shorea Changes are complete and ready for review

shorea:
My apologies I haven't had a chance to take a close look at the code. I've skimmed through it and it looks like the approach is solid. I'll take a closer look tomorrow and provide more meaningful feedback.

aws-java-sdk-core/src/main/java/com/amazonaws/auth/profile/internal/securitytoken/ProfileCredentialsService.java
```diff
@@ -0,0 +1,10 @@
+package com.amazonaws.auth.profile.internal.securitytoken;
+
+import com.amazonaws.auth.AWSCredentials;
+import com.amazonaws.auth.AWSCredentialsProvider;
+
+public interface ProfileCredentialsService {
+    AWSCredentialsProvider assumeRole(AWSCredentials longLivedCredentials, RoleInfo targetRoleInfo);
```
shorea: What do you think about moving the long lived credentials into RoleInfo and having a convenience setter for AWSCredentials that wraps it in a static provider. Could simplify the interface a bit.

aws-java-sdk-core/src/main/java/com/amazonaws/auth/profile/internal/ProfilesConfigFileLoader.java
```diff
@@ -82,43 +85,99 @@
...
+        } else {
+            sourceCredentials = new BasicSessionCredentials(sourceAccessKey, sourceSecretKey, sourceSessionToken);
+        }
+
+        if (roleSessionName == null) {
+            roleSessionName = "aws-sdk-java-" + System.currentTimeMillis();
+        }
+
+        RoleInfo roleInfo = new RoleInfo(roleArn, roleSessionName, externalId);
+        return new Profile(profileName, sourceProfileName, profileCredentialsService.assumeRole(sourceCredentials, roleInfo), roleInfo);
```
shorea: If I'm reading the code correctly it seems we will try and assume the role for every profile we find regardless of whether it's used or not. It's odd that each ProfileCredentialsProvider will parse and load all profiles each time. I think I'd prefer we have a single instance of profile metadata that doesn't try to do anything with the data (i.e. no assuming roles) and when a ProfileCredentialsProvider is created it would reach out to this static registry of profile data (lazily loaded), get the information in the profile and from there do something with it (i.e. create credentials object, assume role, etc). That way we only throw an exception when a customer is using a assume role profile without having STS rather then just having it in their profile file but not actually reference it in the Java SDK (which I imagine would be pretty common since it works for the CLI and they could just be using different profiles for different tools)

aws-java-sdk-core/src/main/java/com/amazonaws/auth/profile/internal/securitytoken/STSProfileCredentialsServiceLoader.java
```diff
@@ -0,0 +1,41 @@
...
+
+import com.amazonaws.AmazonClientException;
+import com.amazonaws.auth.AWSCredentials;
+import com.amazonaws.auth.AWSCredentialsProvider;
+
+/**
+ * Loads <code>com.amazonaws.services.securitytoken.internal.STSProfileCredentialsService</code>
+ * from the STS SDK module, if the module is on the current classpath.
+ */
+public class STSProfileCredentialsServiceLoader implements ProfileCredentialsService {
```
shorea: This looks like it should be a singleton.

aws-java-sdk-core/src/main/java/com/amazonaws/auth/profile/internal/ProfilesConfigFileLoader.java
dump247: It will not attempt to assume the role until the credentials are requested from the Profile object.

Maybe I misunderstood, but previously you said `If we encounter a role based profile and we can't find an appropriate STS implementation then we should fail fast instead of skipping to prevent non deterministic loading of credentials.` I took that to mean that it should fail if any role profile is encountered. I agree that it makes more sense to not fail until the profile credentials are actually requested.

dump247: A static registry of profile metadata seems like a bigger scope than this change.

aws-java-sdk-core/src/main/java/com/amazonaws/auth/profile/internal/securitytoken/STSProfileCredentialsServiceLoader.java
dump247: OK

aws-java-sdk-core/src/main/java/com/amazonaws/auth/profile/internal/securitytoken/ProfileCredentialsService.java
dump247: OK

aws-java-sdk-core/src/main/java/com/amazonaws/auth/profile/internal/ProfilesConfigFileLoader.java
shorea: Ah you're right. I think the naming of the method confused me. assumeRole returns a provider which will not eagerly fetch credentials.

I don't want it to fail if it sees a profile with assume role features but I do want it to fail if a customer attempts to use one without have the STS impl available.

aws-java-sdk-core/src/main/java/com/amazonaws/auth/profile/internal/securitytoken/ProfileCredentialsService.java
```diff
@@ -0,0 +1,10 @@
...
+
+    AWSCredentialsProvider assumeRole(AWSCredentialsProvider longLivedCredentialsProvider, RoleInfo targetRoleInfo);
```
shorea: Consider renaming this to something like getAssumeRoleCredentialsProvider

aws-java-sdk-core/src/main/java/com/amazonaws/auth/profile/internal/ProfilesConfigFileLoader.java
dump247: Ah, I see the confusion. The assumeRole method returns an `AWSCredentialsProvider` which is an `STSAssumeRoleSessionCredentialsProvider`, which doesn't assume the role until the credentials are requested.

dump247:
@shorea Changes made. The class in the STS module should now only be loaded when the profile credentials are requested.

aws-java-sdk-core/src/main/java/com/amazonaws/auth/profile/internal/securitytoken/STSProfileCredentialsServiceProvider.java
```diff
@@ -0,0 +1,64 @@
...
+    private AWSCredentialsProvider getProfileCredentialsProvider() {
+        if (this.profileCredentialsProvider == null) {
+            this.profileCredentialsProvider = getProfileCredentialService().getAssumeRoleCredentialsProvider(roleInfo);
+        }
+
+        return this.profileCredentialsProvider;
+    }
+
+    private static ProfileCredentialsService getProfileCredentialService() {
+        if (STS_SERVICE_CONSTRUCTOR == null) {
```
shorea: Sorry this was more my motiviation in making the Loader a singleton. I'd rather not have to use reflection more than once. We can lazily instantiate the ProfileCredentialsService and cache it in the singleton reference.

aws-java-sdk-core/src/test/resources/resources/profileconfig/ProfileWithRole.tst
```diff
@@ -0,0 +1,7 @@
+[source]
```
shorea: Can you add a test where the source is defined after the role profile? I feel like that would be an easy case to miss if we ever refactored the profile code.

aws-java-sdk-core/src/main/java/com/amazonaws/auth/profile/internal/ProfilesConfigFileLoader.java
```diff
@@ -82,43 +85,103 @@
...
+
+        Map<String, String> sourceProfileProperties = allProfileProperties.get(sourceProfileName);
+
+        if (sourceProfileProperties == null) {
+            throw new AmazonClientException(
+                    String.format("Unable to load source profile [%s]: Source profile not found [%s]",
+                            profileName, sourceProfileName));
+        }
+
+        String sourceAccessKey = sourceProfileProperties.get(Profile.AWS_ACCESS_KEY_ID);
```
shorea: Is there anyway we can just get the source profile by name rather than having to reparse the properties?

shorea:
Had a few more comments and questions. Don't have any major concerns with the code. Can you confirm you are submitting under the Apache 2.0 license.

aws-java-sdk-core/src/test/resources/resources/profileconfig/ProfileWithRole.tst
dump247: Added

aws-java-sdk-core/src/main/java/com/amazonaws/auth/profile/internal/securitytoken/STSProfileCredentialsServiceProvider.java
dump247: Not sure what I was thinking here. I changed to caching the ProfileCredentialsService.

dump247:
Yes, Apache 2.0. I added the header comment to the files I added.

aws-java-sdk-core/src/main/java/com/amazonaws/auth/profile/internal/ProfilesConfigFileLoader.java
dump247: There are a few issues:

1. The source profile may not have been parsed yet, so may not be in the profilesByName map.
2. The source profile can not be another role based profile. That is partly because it is the way botocore implements it and partly because then you could have a chain of role assumptions. It would require some cycle checks and generally be more complex for little gain.

shorea:
Quick update, just waiting on internal approval to get this merged in.

shorea:
Got the green light to merge. We've got a release today but unfortunately won't have time to include this change in it. I expect this to be released this Thursday but that's still tentative.

temujin9:
Unfortunately, @dump247 has moved on to another position, and may be too busy getting settled in to help us address this further.

@shorea Do we need to do any further work from our team to get this merged? It looks like the Travis CI failure was a simple OOM, not a code issue, but I see there are also merge conflicts.

dump247:
@temujin9 I am still on board with this change. The merge conflict is the version number in the sts pom file. For some reason, the aws sdk references the previous release version for dependencies between modules. Not sure why that is, but it results in a conflict whenever a new version is released. The rebase is straight forward.

@shorea Any reason for the delay in merging? It sounded like merging was a given. Has anything changed?

shorea:
Sorry I've staged this for release internally. It's going out today. Merging PR now.

temujin9:
Awesome, thank you both.

PR #643, opened 2016-02-25
Add setMultipartUploadThreshold(int) for Hadoop 2.7 S3A FS

romi-totango:
replaced previous PR https://github.com/aws/aws-sdk-java/pull/637

reverting backward incompatibility introduced in https://github.com/aws/aws-sdk-java/pull/201

patch released under Apache 2.0 license

### DIFF
--- a/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/transfer/TransferManagerConfiguration.java
+++ b/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/transfer/TransferManagerConfiguration.java
@@ -14,6 +14,8 @@
  */
 package com.amazonaws.services.s3.transfer;
 
+import java.lang.Deprecated;
+
 import static com.amazonaws.services.s3.internal.Constants.*;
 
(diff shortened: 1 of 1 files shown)


romi-totango:
@breedloj

breedloj:
Thanks for the update.  The updated fix should be going out with today's release!

anshbansal:
@breedloj I tried with 1.11.126 and it fails. Was the method removed in 1.11.+ and only works in 1.10.+?

ebuildy:
HEllo @anshbansal , this fix is here from tag 1.10.60:

https://github.com/aws/aws-sdk-java/blob/1.10.60/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/transfer/TransferManagerConfiguration.java

It's still here in the version you are talking about: https://github.com/aws/aws-sdk-java/blob/1.11.126/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/transfer/TransferManagerConfiguration.java#L248

PR #650, opened 2016-03-08
Fix subscribeQueue issue with empty policy

chidoli:
Fix the issue that subscribeQueue with extendPolicy cannot make subscription when there is no existing policy.

### DIFF
--- a/aws-java-sdk-sns/src/main/java/com/amazonaws/services/sns/util/Topics.java
+++ b/aws-java-sdk-sns/src/main/java/com/amazonaws/services/sns/util/Topics.java
@@ -186,8 +186,9 @@ public class Topics {
         Map<String, String> sqsAttrs = sqs.getQueueAttributes(sqsQueueUrl, sqsAttrNames).getAttributes();
         String sqsQueueArn = sqsAttrs.get(QueueAttributeName.QueueArn.toString());
 
-        Policy policy = extendPolicy ? Policy.fromJson(
-                sqsAttrs.get(QueueAttributeName.Policy.toString())) : new Policy();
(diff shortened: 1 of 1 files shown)


aws-java-sdk-sns/src/main/java/com/amazonaws/services/sns/util/Topics.java
```diff
@@ -186,8 +186,9 @@ public static String subscribeQueue(AmazonSNS sns, AmazonSQS sqs, String snsTopi
         Map<String, String> sqsAttrs = sqs.getQueueAttributes(sqsQueueUrl, sqsAttrNames).getAttributes();
         String sqsQueueArn = sqsAttrs.get(QueueAttributeName.QueueArn.toString());
 
-        Policy policy = extendPolicy ? Policy.fromJson(
-                sqsAttrs.get(QueueAttributeName.Policy.toString())) : new Policy();
+        String policyJson = sqsAttrs.get(QueueAttributeName.Policy.toString());
+        Policy policy = extendPolicy || policyJson == null || policyJson.length() == 0
```
breedloj: Maybe I'm missing something here but don't you only want to extend the policy if the flag is set, policyJson is NOT null and NOT empty?

chidoli: I have several environments and I guess I got confused. I will correct it. Thanks a lot.

breedloj:
Change looks good.  Can you confirm you are submitting this under the Apache 2.0 license?

chidoli:
Yes

breedloj:
This has been staged for release and should be available shortly.

PR #656, opened 2016-03-14
Sys Env Var HTTPS?_PROXY support in com.amazonaws.ClientConfiguration#@ClientConfiguration()

norbertkorodi:
The env variables HTTPS?_PROXY can be used to obtain proxy setting infos in com.amazonaws.ClientConfiguration#@ClientConfiguration().
Default is to use them if the appropriate Systemproperty is not available.


zhangzhx:
Hey @norbertkorodi, we don't want to make the ClientConfiguration class super complicated. It looks like you want to respect the env variable to set proxy and related properties. We even don't want to respect the system properties for proxy related variables, as the HttpClientBuilder will respect those properties anyway (see following link). So we don't want to further open the doors for setting the configurations.

http://hc.apache.org/httpcomponents-client-ga/httpclient/apidocs/org/apache/http/impl/client/HttpClientBuilder.html

By saying that, we will not merge the change. But thanks for the contribution.

norbertkorodi:
hi @zhangzhx,

Thnx for looking into the pr. We assumed that it is very useful to use the environment variable and not the system properties. If you'd like the feature sometime in the future and would like some refactoring, pls let me know.

Br,
Norbert

hoegertn:
@zhangzhx  So can you elaborate on how I can get an application using the SDK to use a proxy set via environment variables if I do not have access to the system properties or the code? The configuration of proxy settings with env variables is quite common and I do not see why the SDK should not use them.

PR #670, opened 2016-03-29
"connection" header should be removed as it is changed to "keep-alive…

ttwd80:
…" regardless of the value sent


breedloj:
Hey, thanks for the PR!  As a matter of fact we just got a PR yesterday that should achieve the same thing: https://github.com/aws/aws-sdk-java/pull/669  This is being merged in for our next scheduled release (should be today).  Please let me know if there is still something additionally you'd like to see changed.

Thanks!

breedloj:
Ah, looking at this closer this looks like a more comprehensive change to the signer vs. the other PR which only affects S3.  Let me check and see if it makes sense to include both of these changes.

shorea:
Can you confirm you are submitting under Apache 2.0?

ttwd80:
Yes. Submitting under Apache 2.0.

shorea:
Thanks for the contribution. This will be included in our release later today.

PR #671, opened 2016-03-29
Add missing queue attribute name and optimize the whole enumeration

mway-dmueller:


shorea:
This is a generated class so we'll have to make changes in the code generator.

zhangzhx:
This has been released. See 

https://github.com/aws/aws-sdk-java/blob/master/aws-java-sdk-sqs/src/main/java/com/amazonaws/services/sqs/model/QueueAttributeName.java#L57-L70

mway-dmueller:
OK, fromValue() method was optimized, but enum type QueueAttributeName.All is still missing.

PR #680, opened 2016-04-07
Fix labels of credentials to match ~/.aws/credentials

petitviolet:
The command `aws configure` generates ~/.aws/credentials file.
And labels of credential values are below.
- `aws_access_key_id`
- `aws_secret_access_key`


manikandanrs:
The file generated by the CLI is the profile credentials file. You will need to use the ProfileCredentialsProvider and not PropertiesCredentials.

ProfileCredentialsProvider looks for the same access key and secret key properties as the CLI.

petitviolet:
Thanks a lot!
I have just wanted to use directly the credentials generated by `aws configure`, however I did not know ProfileCredentialsProvider.

PR #682, opened 2016-04-07
Allow overriding the httpclient/httpcore versions

atokhy:
We would like to be able to override the httpcomponents versions bundled by the AWS SDK through the command line.

mvn package -Dhttpcore.version=4.3.3 -Dhttpclient.version=4.3.6

This way we can independently update the httpclient/httpcore versions.

### DIFF
--- a/aws-java-sdk-core/pom.xml
+++ b/aws-java-sdk-core/pom.xml
@@ -21,7 +21,12 @@
     <dependency>
       <groupId>org.apache.httpcomponents</groupId>
       <artifactId>httpclient</artifactId>
-      <version>4.3.6</version>
+      <version>${httpclient.version}</version>
+    </dependency>
+    <dependency>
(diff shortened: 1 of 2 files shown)


aws-java-sdk-core/pom.xml
```diff
@@ -21,7 +21,12 @@
     <dependency>
       <groupId>org.apache.httpcomponents</groupId>
       <artifactId>httpclient</artifactId>
-      <version>4.3.6</version>
+      <version>${httpclient.version}</version>
```
manikandanrs: apache.httpclient.version ?

atokhy: I was thinking maybe:

httpcomponents.httpclient.version

https://hc.apache.org/

To disambiguate from the older Apache httpclient.

PR #683, opened 2016-04-07
Allow overriding the httpclient/httpcore versions

atokhy:
We would like to be able to override the httpcomponents versions bundled by the AWS SDK through the command line.

mvn package -Dhttpcomponents.httpcore.version=4.3.3 -Dhttpcomponents.httpclient.version=4.3.6

This way we can independently update the httpclient/httpcore versions.

### DIFF
--- a/aws-java-sdk-core/pom.xml
+++ b/aws-java-sdk-core/pom.xml
@@ -21,7 +21,12 @@
     <dependency>
       <groupId>org.apache.httpcomponents</groupId>
       <artifactId>httpclient</artifactId>
-      <version>4.3.6</version>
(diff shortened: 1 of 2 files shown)


atokhy:
The build above appears to have failed due to an OOM.

manikandanrs:
Thanks for the pull request. Can you confirm if you are submitting under Apache 2.0 license ?

atokhy:
Yes, this is being submitted under the Apache 2.0 license.

manikandanrs:
Thanks for the PR. This will be available as part of our next release.

PR #686, opened 2016-04-13
Add CloudFrontCookieSigner static class to aws-java-sdk-cloudfront

handokochen:
Implement the Java version of CloudFront Signed Cookies.

### DIFF
--- a/aws-java-sdk-cloudfront/src/main/java/com/amazonaws/services/cloudfront/CloudFrontCookieSigner.java
+++ b/aws-java-sdk-cloudfront/src/main/java/com/amazonaws/services/cloudfront/CloudFrontCookieSigner.java
@@ -51,24 +51,30 @@ import static java.util.concurrent.TimeUnit.MILLISECONDS;
  * Date expiresOn = DateUtils.parseISO8601Date("2011-11-14T22:20:00.000Z");
  * String ipRange = "192.168.0.1/24";
  *
- * CookiesForCannedPolicy cookies = CloudFrontCookieSigner.getCookiesForCannedPolicy(
+ * CookiesForCannedPolicy cookies1 = CloudFrontCookieSigner.getCookiesForCannedPolicy(
  *              protocol, distributionDomain, privateKeyFile,
  *              s3ObjectKey, keyPairId, activeFrom);
  *
- * CookiesForCustomPolicy url2 = CloudFrontCookieSigner.getCookiesForCannedPolicy(
+ * CookiesForCustomPolicy cookies2 = CloudFrontCookieSigner.getCookiesForCustomPolicy(
  *              protocol, distributionDomain, privateKeyFile,
  *              s3ObjectKey, keyPairId, activeFrom,
  *              expiresOn, ipRange);
+ *
+ * // Converting to javax.servlet.http.Cookie example
+ * List<Cookie> cookies = new ArrayList<>();
+ * for (Map.Entry<String, String> cookie : cookies1.entrySet()) {
+ *  cookies.add(new Cookie(cookie.getKey(), cookie.getValue()));
+ * }
  * </pre>
  *
  * @see DateUtils
  */
 public enum CloudFrontCookieSigner {
     ;
-    private static final String expiresKey = "CloudFront-Expires";
-    private static final String signatureKey = "CloudFront-Signature";
-    private static final String policyKey = "CloudFront-Policy";
-    private static final String keyPairIdKey = "CloudFront-Key-Pair-Id";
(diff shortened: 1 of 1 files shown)


handokochen:
The build above appears to have failed due to an OOM.

zhangzhx:
Could you provide your test cases as well if any?

handokochen:
There are no test cases because it needs a real AWS account with shared CloudFront Encryption Key PEM file to work. I've tested it on a full stack of S3 origin + CloudFront Distribution with Signed Cookie in my project.

I'm following the documentation in the comment section of com.amazonaws.services.cloudfront.CloudFrontUrlSigner class for code example and how to store the cookie using javax.servlet.http.Cookie (line 54-67). It's only a comment because servlet-api isn't part of the pom dependency.

aws-java-sdk-cloudfront/src/main/java/com/amazonaws/services/cloudfront/CloudFrontCookieSigner.java
```diff
@@ -0,0 +1,525 @@
...
+        }
+    }
+
+    /**
+     * Contains the cookies used to access restricted content from
+     * CloudFront using a canned policy
+     */
+    public static class CookiesForCannedPolicy extends SignedCookies {
+        // Date and time in Unix time format (in seconds) and Coordinated Universal Time (UTC).
+        public Map<String, String> expires;
```
zhangzhx: This should be a <key,value> pair instead of a Map.

aws-java-sdk-cloudfront/src/main/java/com/amazonaws/services/cloudfront/CloudFrontCookieSigner.java
```diff
@@ -0,0 +1,525 @@
...
+        }
+    }
+
+    /**
+     * Contains the cookies used to access restricted content from
+     * CloudFront using a custom policy.
+     */
+    public static class CookiesForCustomPolicy extends SignedCookies {
+        // Base64 encoded version of the custom policy.
+        public Map<String, String> policy;
```
zhangzhx: Same here and the Maps in SignedCookies.

aws-java-sdk-cloudfront/src/main/java/com/amazonaws/services/cloudfront/CloudFrontCookieSigner.java
```diff
@@ -0,0 +1,525 @@
...
+             ;
+    }
+
+    /**
+     * Creates a private key from the file given, either in RSA private key
+     * pkcs8 (.der) format. Other formats, including .pem, will cause an exception to
+     * be thrown.
+     */
+    static PrivateKey loadPrivateKey(File privateKeyFile) throws InvalidKeySpecException, IOException {
+        if ( StringUtils.lowerCase(privateKeyFile.getAbsolutePath()).endsWith(".der") ) {
```
zhangzhx: Why does it not support pem while presigned URL does?

aws-java-sdk-cloudfront/src/main/java/com/amazonaws/services/cloudfront/CloudFrontCookieSigner.java
handokochen: Updated.

aws-java-sdk-cloudfront/src/main/java/com/amazonaws/services/cloudfront/CloudFrontCookieSigner.java
handokochen: Updated.

zhangzhx:
This change looks good to me. As this CookieSigner class is sharing a lot duplicate code with UrlSigner, I will refactor a little bit, but not changing the API functionality. Once it gets approved from our team, it'll be published from next release.

zhangzhx:
Hi, we'll still use an API similar to .net for signed cookies like the following.

```java
HttpClient client = new DefaultHttpClient();
HttpGet httpGet = new HttpGet(SignerUtils.generateResourcePath(protocol, distributionDomain, resourcePath));
httpGet.addHeader("Cookie", cookies.getExpires().getKey() + "=" + cookies.getExpires().getValue());
httpGet.addHeader("Cookie", cookies.getSignature().getKey() + "=" + cookies.getSignature().getValue());
httpGet.addHeader("Cookie", cookies.getKeyPairId().getKey() + "=" + cookies.getKeyPairId().getValue());
HttpResponse response = client.execute(httpGet);
```

So, instead of merging the PR, I will just release the change in the API style above. The next release is on Tuesday on which you will see the update by EOD.

PR #696, opened 2016-04-27
Fix SignerUtils.buildCustomPolicy parameter names

mhurne:
Currently, the value of the `activeFrom` parameter to `SignerUtils#buildCustomPolicy` is used to populate the `DateLessThan` property of the generated policy and the `expiresOn` parameter is used to populate the `DateGreaterThan` property. Callers will be surprised to find that their `expiresOn` parameter gets used to determine when the signature starts being valid while their `activeFrom` parameter is used to determine when the signature expires. This pull request renames the parameters such that `activeFrom` is used to populate the `DateGreaterThan` property in the policy while `expiresOn` is used to populate the `DateLessThan` property, which is what callers would expect.

Conveniently, all of the callers to the modified method within the library itself were passing their `expiresOn` for the `activeFrom` parameter, and their `activeFrom` for the `expiresOn` parameter, so none of them needed to be modified to accommodate this change.

### DIFF
--- a/aws-java-sdk-cloudfront/src/main/java/com/amazonaws/services/cloudfront/util/SignerUtils.java
+++ b/aws-java-sdk-cloudfront/src/main/java/com/amazonaws/services/cloudfront/util/SignerUtils.java
@@ -54,22 +54,22 @@ public class SignerUtils {
      * Returns a custom policy for the given parameters.
      */
     public static String buildCustomPolicy(String resourcePath,
-            Date activeFrom, Date expiresOn, String ipAddress) {
+            Date expiresOn, Date activeFrom, String ipAddress) {
         return "{\"Statement\": [{"
                 + "\"Resource\":\""
                 + resourcePath
                 + "\""
                 + ",\"Condition\":{"
                 + "\"DateLessThan\":{\"AWS:EpochTime\":"
-                + MILLISECONDS.toSeconds(activeFrom.getTime())
+                + MILLISECONDS.toSeconds(expiresOn.getTime())
                 + "}"
                 + ",\"IpAddress\":{\"AWS:SourceIp\":\""
                 + ipAddress
                 + "\"}"
-                + (expiresOn == null
+                + (activeFrom == null
                    ? ""
                    : ",\"DateGreaterThan\":{\"AWS:EpochTime\":"
-                     + MILLISECONDS.toSeconds(expiresOn.getTime()) + "}"
+                     + MILLISECONDS.toSeconds(activeFrom.getTime()) + "}"
                   )
                 + "}}]}";
     }


varunnvs92:
Hi,
Thank you for the pull request. Can you provide us the following details:
1) Short description of the change. What is the reason behind interchanging the parameters in the method signature.
2) Are you submitting it under Apache 2.0 license?

Thanks!

mhurne:
I've updated the originally empty description with details about the change, and I am indeed submitting the changes under the Apache 2.0 license. Thanks!

varunnvs92:
Hi @mhurne,
The change in the method signature which interchanges activeFrom and expiresOn is a breaking change to current customers. So we should keep the signature as it is.
But we can change the method body as you pointed out in the PR. I can make the changes in our code or you can update the PR with only the changes in method body and send it again to us.

Thanks for the PR and pointing out the issue.

mhurne:
@varunnvs92 I disagree - my changes are non-breaking. If the method body is changed to swap the usages of the parameters but the names of parameters are not changed, that _will_ be a breaking change.

All I did was rename the parameters; I didn't change how they are used. Both before and after my changes, the first `Date` was and is used to populate the `DateLessThan` in the policy and the second `Date` was and is used to populate the `DateGreaterThan` in the policy.

If there was test coverage of the existing code, I expect that the test would continue to pass with these changes without modification of the test.

varunnvs92:
@mhurne 
Assume a customer who wants to set the activeFrom value as 1000 and expiresOn value as 2000. Currently he would be using this (String resourcePath, Date activeFrom, Date expiresOn, String ipAddress) method as (resource, 1000, 2000, ip). Now if we change the method signature as (String resourcePath, Date expiresOn, Date activeFrom, String ipAddress), the customer's code which is (resource, 1000, 2000, ip) seems like he is using 1000 for expiresOn and 2000 for activeFrom which is breaking change/confuses the customer.

I know that the your implementation works good for new customers but may confuse customers already using this function. I agree that the current code has bug and doesn't work as expected. So fixing the body should be good.

mhurne:
@varunnvs92 The thing is, the meaning of the current `activeFrom` parameter is actually `expiresOn` and vice-versa. My changes don't change the way the parameters are used, it just fixes the names. In your example, `DateLessThan` will be based on `1000` and `DateGreaterThan` will be based on `2000`. That doesn't make any sense, and is exactly the issue I ran into that led to these changes...I was able to get the method to behave the way I wanted by pretending that `activeFrom` was actually named `expiresOn` and vice-versa. This changeset just fixes the names. If the parameters didn't have names, this changeset wouldn't be needed. I don't know how else to clarify this. I literally did a "Refactor->Rename" in IntelliJ for each parameter. I didn't simply swap the parameters in the signature, I renamed them in both the signature and the method body to ensure that their behavior did not change.

Perhaps we should add a test around the existing implementation, merge that in to master first, and then merge that test in to my fork/branch. If we do that, I'm confident the test will continue to pass without needing to change here.

mhurne:
Even existing calls within the SDK call the method as if the params are `expiresOn` and then `activeFrom`. I didn't have to update those calls as part of these changes because they behavior those calls expect was not changed.

varunnvs92:
@mhurne 
Cool! Everything looks good. I will merge this in upon internal approval. Hopefully it should be included in today's release. If not, it will be included in Thursday's release.
Thank you for your contribution 👍

mhurne:
@varunnvs92 Thanks!

Issue #660, opened 2016-03-17
AmazonCloudSearchDomainClient does not expose a way to set "search" requests to use HTTP POST


humanzz:
As per the documentation at http://docs.aws.amazon.com/cloudsearch/latest/developerguide/submitting-search-requests.html, HTTP GET requests have a size limit of 8190 bytes. To get around that, one might want to submit search requests via an HTTP POST. However, the AmazonCloudSearchDomainClient class does not expose any way to change the default HTTP GET to become a POST.

Ideally, setting the HTTP method for the search HTTP request should be exposed.

zhangzhx:
We've had discussed this issue in our team meeting. Unfortunately, we are not planing to support both POST and GET in the SDK. We are relying on the service model to generate our SDK code. It's not a easy work to support both http methods simply for this operation. A workaround is perhaps try to shrink the size.

dagnir:
Hi @humanzz, in case you were still interested in this feature, this is now fixed as of [1.11.32](https://aws.amazon.com/releasenotes/Java/0627162592215831) and all search requests now use `POST`. Thank for reporting this!

PR #697, opened 2016-04-27
Exposing a way to set search requests to use HTTP POST for AmazonClou…

emata-socialware:
Hi all,

We have been working with AWS CloudSearch for a while and it has been a good experience, nonetheless we found that there is a limitation for the search request size when doing a GET request.

There are few search requests on our end that are bigger than 8190 bytes. We have reviewed them and the complexity is not that hard, so, we would like to perform a POST request for these corner cases as CloudSearch documentation mention: "... The maximum size of a search request submitted via GET is 8190 bytes, including the HTTP method, URI, and protocol version. You can submit larger requests using HTTP POST; however, keep in mind that large, complex requests take longer to process and are more likely to time out. For more information..."

However we found that the Java AWS SDK does not allow the developer to send a POST request for larger requests. So, we ended-up forking the official SDK repository in order to implement that functionality and eventually share that with the rest of the community since we know they are also requesting such functionality.

Note that the code changes were intended to fulfill the initial requirements at https://github.com/aws/aws-sdk-java/issues/660 and also to be compliant with official CloudSearch documentation.

Best regards!

P.S: Please note that the fix will apply not only to AmazonCloudSearchDomainClient but also to AmazonCloudSearchDomainAsyncClient

### DIFF
--- a/aws-java-sdk-cloudsearch/src/main/java/com/amazonaws/services/cloudsearchdomain/AmazonCloudSearchDomainClient.java
+++ b/aws-java-sdk-cloudsearch/src/main/java/com/amazonaws/services/cloudsearchdomain/AmazonCloudSearchDomainClient.java
@@ -323,8 +323,12 @@ public class AmazonCloudSearchDomainClient extends AmazonWebServiceClient
         try {
             awsRequestMetrics.startEvent(Field.RequestMarshallTime);
             try {
-                request = new SearchRequestMarshaller().marshall(super
-                        .beforeMarshalling(searchRequest));
+                // Check if user wants to perform a large request using HTTP POST. Use HTTP GET otherwise.
+                if (searchRequest.getHttpMethod() == HttpMethodName.POST) {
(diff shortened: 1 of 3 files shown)


varunnvs92:
Hi @emata-socialware,
Thank you for contributing the code. Unfortunately we cannot merge this change as the AmazonCloudSearchDomainClient is auto generated from the service model(see [this](https://github.com/aws/aws-sdk-java/blob/master/aws-java-sdk-cloudsearch/src/main/resources/models/cloudsearchdomain-2013-01-01-model.json#L29) for model). If Java team decides to support POST without modifications from service team's model, we need to make customizations to code generator which changes all the auto generated Java clients. This may cause issues with many current customers and may break their code.

I will raise an issue with the service team to see if they can change the model to add POST method for the Search operation.  I will update you once I receive a response from the service team. 

Thank you again for your effort and contributing to the community.

dagnir:
Hi @emata-socialware, this issue has now been fixed as of version [1.11.32](https://aws.amazon.com/releasenotes/Java/0627162592215831); all `SearchRequest`s now use `POST` instead of `GET`. Closing this PR.

PR #718, opened 2016-05-20
Remove unnecessary use of JAXB

rjernst:
In the latest Java 9 ea build (118), the default modules for compile and
runtime were changed (see
http://mail.openjdk.java.net/pipermail/jdk9-dev/2016-May/004309.html).
In particular, java.xml.bind was removed. There is exactly one use of it
here: using DatatypeConverter to encode a base64 string. However, the
Base64Codec used by the Base64 class here already supports encoding. In
version 1.8.3 of the sdk, this was mysteriously changed to use jaxb.

This PR changes the method back to using Base64Codec, as it was before
1230cff. There is no need to use jaxb here, especially since it will
simply not work with java 9 without requiring extra arguments in the
java command line.

### DIFF
--- a/aws-java-sdk-core/src/main/java/com/amazonaws/util/Base64.java
+++ b/aws-java-sdk-core/src/main/java/com/amazonaws/util/Base64.java
@@ -14,10 +14,6 @@
  */
 package com.amazonaws.util;
 
-import javax.xml.bind.DatatypeConverter;
-
-import com.amazonaws.log.InternalLogFactory;
-
 /**
  * A Base 64 codec API.
  *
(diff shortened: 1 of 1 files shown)


zhangzhx:
This change looks good to me. I will make this change internally and once approved by the team, I will merge this PR.

zhangzhx:
We've discussed about this in the team, and we want to hold off this change as Java 9 is still a long time away. We also want to conduct more performance tests to see the potential performance impact for removing this JAXB encoding. See the following link for the base64 codec performance tests we did before.

https://java.awsblog.com/post/Tx27TR56RDBC58U/Follow-up-on-Base64-Codec-Performance

yegeniy:
Should this be reopened and prioritized?

I've run into a JAXB issue from this when running on java v9.0.1:

```
java.lang.NoClassDefFoundError: javax/xml/bind/JAXBException
        at com.amazonaws.services.s3.AmazonS3Client.putObject(AmazonS3Client.java:1710) ~
        at com.amazonaws.services.s3.AmazonS3Client.putObject(AmazonS3Client.java:3264) ~
        ....
Caused by: java.lang.ClassNotFoundException: javax.xml.bind.JAXBException
        at jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:582) ~[?:?]
        at jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:185) ~[?:?]
        at java.lang.ClassLoader.loadClass(ClassLoader.java:496) ~[?:?]
        ... 22 more
```

yegeniy:
In the meantime, here's one workaround to get it working: https://stackoverflow.com/a/43601685/2916086

> Provide the `--add-modules java.xml.bind` flag to both `javac` and `java`. If you're in maven, leverage the [maven-compiler-plugin's arg](https://maven.apache.org/plugins/maven-compiler-plugin/examples/pass-compiler-arguments.html):

```
  <build>
    <plugins>
      <plugin>
        <groupId>org.apache.maven.plugins</groupId>
        <artifactId>maven-compiler-plugin</artifactId>
        <version>3.7.0</version>
        <configuration>
          <compilerArgs>
            <!-- This helps aws's S3 module to work with Java 9: -->
            <!-- https://docs.oracle.com/javase/9/migrate/#GUID-F640FA9D-FB66-4D85-AD2B-D931174C09A3 -->
            <arg>--add-modules=java.xml.bind</arg>
...
```

PR #723, opened 2016-05-30
Add an overload to TransferManager.copy(...) with a 'metadata' parameter

spew:
When performing a cross region copy it is desirable to use an AmazonS3Client for the destination region. If one uses the source region as their S3 client then an AmazonS3Exception is thrown telling the client to re-direct to the correct endpoint. This is handled by S3V4AuthErrorRetryStrategy but it is slow and causes a constant stream of error messages to be printed to the console about the redirection:

  "please use region-specific endpoint to access ... buckets located in regions that require V4 signing."

Today, the TransferManager.copy(...) method requires that the source region S3 be used because of a call to fetch the ObjectMetadata in the copy  method. This getObjectMetadata request must be made with an S3Client pointing at the source region. This commit adds an overload which allows the user to pass in the metadata and thus use a TransferManager with a destination S3 client.

### DIFF
--- a/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/transfer/TransferManager.java
+++ b/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/transfer/TransferManager.java
@@ -1614,6 +1614,57 @@ public class TransferManager {
      */
 
     public Copy copy(final CopyObjectRequest copyObjectRequest,
+                     final TransferStateChangeListener stateChangeListener)
+        throws AmazonServiceException, AmazonClientException {
+
+        GetObjectMetadataRequest getObjectMetadataRequest =
+            new GetObjectMetadataRequest(
+                copyObjectRequest.getSourceBucketName(),
+                copyObjectRequest.getSourceKey())
+                .withSSECustomerKey(copyObjectRequest.getSourceSSECustomerKey());
+        ObjectMetadata metadata = s3.getObjectMetadata(getObjectMetadataRequest);
+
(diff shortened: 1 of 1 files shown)


shorea:
How do you feel about an overload that takes an instance of AmazonS3 instead for any service calls that need to be made to the destination region?

spew:
shorea: that works too! So the actual "AmazonS3.copyObject(...)" request would be made to the destination in that case but the objectMetadata would be retrieved from the source region using the default s3 client?

shorea:
Good question. Maybe it makes more sense to pass in the client of the source region and use the transfer manager's client for the actual copy object. Since the exposed API is copy I think I'd expect it to use for TF's client for the copy. Open for negotiation though, I think either approach would be fine.

spew:
No problem with either, I agree passing in the source client for the metadata operation is better (your second suggestion).

manikandanrs:
@spew @shorea 

The [GetObjectMetadata](https://github.com/aws/aws-sdk-java/blob/master/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/AmazonS3Client.java#L997) API on the AmazonS3Client automatically resolves the region of bucket by making a HEAD calls and caching the region of the bucket for future calls. Did you try initializing the TransferManager with the destination and see if it works ?

spew:
Destination works, but it works poorly. Each copy request results in a WARN being printed out in S3V4AuthErrorRetryStrategy it also doubles the number of API calls and leads to throttling.

spew:
Closing this PR and opening a new one with the source client approach.

PR #728, opened 2016-06-02
Daemonizing the sts thread used for the credential refresh

jmccarty3:
This commit demonizes the threads used by the RefreshableTask object to refresh STS credentials as needed. Without this, any clients that use STS credentials and are held (possibly in other daemon threads) past the refresh timer will cause the JVM to hang on destruction.


sstarcher:
We currently can not upgrade past 1.10.64 as 1.10.65 introduces this issue

shorea:
Looks good to me. Can you confirm in the comments you are submitting under apache 2.0?

jmccarty3:
@shorea Yes I can confirm I am submitting under apache 2.0

jmccarty3:
@shorea Is there anything else you need from me with this?

shorea:
This will be included in the next release of the SDK, tentatively scheduled for tomorrow.

PR #732, opened 2016-06-10
S3 Pre-sign signature calculation changed?

ttwd80:
for pre-sign requests, it seems that S3 will use "UNSIGNED-PAYLOAD".
calculating the actual payload and sending them to S3 will result in an error message.
S3 will report that the "String-To-Sign" value uses "UNSIGNED-PAYLOAD".
AWS PHP SDK has a function called "getPresignedPayload" with returns this constant
This looks new as the php code change looks recent.
Ref:
https://github.com/aws/aws-sdk-php/pull/998
https://github.com/aws/aws-sdk-php/blob/master/src/Signature/S3SignatureV4.php

### DIFF
--- a/aws-java-sdk-core/src/main/java/com/amazonaws/auth/AWS4Signer.java
+++ b/aws-java-sdk-core/src/main/java/com/amazonaws/auth/AWS4Signer.java
@@ -235,7 +235,7 @@ public class AWS4Signer extends AbstractAWSSigner implements
         addPreSignInformationToRequest(request, sanitizedCredentials,
                 signerRequestParams, timeStamp, expirationInSeconds);
 
-        final String contentSha256 = calculateContentHashPresign(request);
+        final String contentSha256 = "UNSIGNED-PAYLOAD";
 
         final String canonicalRequest = createCanonicalRequest(request,
                 contentSha256);


manikandanrs:
Java SDK uses "UNSIGNED-PAYLOAD" . [Here](https://github.com/aws/aws-sdk-java/blob/master/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/internal/AWSS3V4Signer.java#L71) is the code.

ttwd80:
```
GeneratePresignedUrlRequest generatePresignedUrlRequest = new GeneratePresignedUrlRequest(bucketName,
				objectKey);
		generatePresignedUrlRequest.setMethod(HttpMethod.GET); // Default.
		generatePresignedUrlRequest.setExpiration(expiration);
		AmazonS3 amazonS3 = new AmazonS3Client(credentials, clientConfiguration);
		URL s = amazonS3.generatePresignedUrl(generatePresignedUrlRequest);
This code, AmazonS3Client.generatePresignedUrl(), calls com.amazonaws.auth.AWS4Signer.presignRequest(), this in turn calls 
calculateContentHashPresign(), which calls calculateContentHash(), which does the actual hash calculation.
com.amazonaws.services.s3.internal.AWSS3V4Signer is not involved when AmazonS3Client.generatePresignedUrl() is called and therefore the code generated will NOT use "UNSIGNED-PAYLOAD"
Maybe I am missing something in the way I call the AmazonS3Client.generatePresignedUrl() function?

```

ttwd80:
i had this, clientConfiguration.setSignerOverride("AWS4SignerType");
closing the issue.

PR #751, opened 2016-06-21
Removed superfluous word from error message

csapir:
getter.getDeclaringClass() prints: "class theClassTheErrorOrginatedFrom"
Therefore, the word 'class' is repeated & confusing.


breedloj:
Thanks for the PR!  This change seems reasonable to me.  Will make sure it gets pushed out with our next release!

csapir:
thanks!

PR #753, opened 2016-06-22
Fix race condition bug inside of AWS4Signer.presignRequest() causing …

flofreud:
…randomly wrong signed URLs

The `AWS4Signer.presignRequest()`-method, which is used to generate presigned S3 URLs, is not always signing correctly. 

Sometimes the presigned URL is wrong because the necessary AWS timestamp is generated twice based on a `System.currentTimeMillis()`-call. If the two calls don't happen within the same second the signingString is generated with a different timestamp as the one added to the signed request. Therefore S3 checks the signature against the wrong timestamp. 

The expected behavior (see http://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-query-string-auth.html) defines `X-Amz-Date` as 
> The date in ISO 8601 format, for example, 20130721T201207Z. This value must match the date value used to calculate the signature.. 

The `sign()`-method uses the same timestamp for both as expected (see https://github.com/flofreud/aws-sdk-java/blob/476490e1a1017dc664c90ad66122a1a645a3dc46/aws-java-sdk-core/src/main/java/com/amazonaws/auth/AWS4Signer.java#L174).

This PR fix the issue and adds a basic test case for the `presign()`-method.

### DIFF
--- a/aws-java-sdk-core/src/main/java/com/amazonaws/auth/AWS4Signer.java
+++ b/aws-java-sdk-core/src/main/java/com/amazonaws/auth/AWS4Signer.java
@@ -229,8 +229,8 @@ public class AWS4Signer extends AbstractAWSSigner implements
                 AWS4_SIGNING_ALGORITHM);
 
         // Add the important parameters for v4 signing
-        final String timeStamp = AWS4SignerUtils.formatTimestamp(System
-                .currentTimeMillis());
+        final String timeStamp = signerRequestParams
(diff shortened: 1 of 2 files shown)


shorea:
Looks good to me. Can you confirm you are submitting under Apache 2.0?

flofreud:
Yes, the submission is under Apache License 2.0

shorea:
Merged, thanks for the fix!

PR #758, opened 2016-06-27
Added mapping for 'us-east-1' to the S3 Region.fromValue method

nterry:
This is a logical default as US_Standard is us-east-1l

This will also help avoid confusion as other APIs return 'us-east-1'

### DIFF
--- a/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/model/Region.java
+++ b/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/model/Region.java
@@ -232,8 +232,9 @@ public enum Region {
 
     /**
      * Returns the Amazon S3 Region enumeration value representing the specified Amazon
-     * S3 Region ID string. If specified string doesn't map to a known Amazon S3
-     * Region, then an <code>IllegalArgumentException</code> is thrown.
+     * S3 Region ID string. If specified string is 'null', 'us-east-1', or 'US',
+     * then {@link #US_Standard} is returned, and if specified string doesn't map
+     * to a known Amazon S3 Region, then an <code>IllegalArgumentException</code> is thrown.
      *
      * @param s3RegionId
      *            The Amazon S3 region ID string.
@@ -245,9 +246,8 @@ public enum Region {
      *             If the specified value does not map to one of the known
      *             Amazon S3 regions.
      */
-    public static Region fromValue(final String s3RegionId) throws IllegalArgumentException
-    {
-        if (s3RegionId == null || s3RegionId.equals("US"))
+    public static Region fromValue(final String s3RegionId) throws IllegalArgumentException {
+        if (null == s3RegionId || "US".equals(s3RegionId) || "us-east-1".equals(s3RegionId))
             return Region.US_Standard;
         for (Region region : Region.values()) {
             List<String> regionIds = region.regionIds;


nterry:
Argh!!! Stupid OOM errors.... I cant build fully locally either....

zhangzhx:
Can you remove dependency-reduced-pom.xml changes. Those don't seem to be related to the Region.fromValue change.

nterry:
@zhangzhx Sure.

zhangzhx:
Can you tell us what the intention you made this change? What is the use case you want this change?

This is a breaking change since providing 'us-east-1' will throw an Exception beforehand, so we will not consider merging this change.

nterry:
Its an annoyance that 'us-east-1' isn't recognized because thats the region of 'US_Standard' in every other service... 

Also, thats what a major version rev is for.... http://semver.org/

I dont mind if this went in the next major release

zhangzhx:
It is pretty confusing here using us-east-1 to refer to US_Standard. S3 is a a little different from other services that the US_Standard (also us-east-1, see following link) region is indeed a global region, it is really misleading. 

http://docs.aws.amazon.com/general/latest/gr/rande.html#s3_region

So I don't recommend setting in this way unless S3 someday officially moves to the right trail, referring us-east-1 to an endpoint (something like s3.us-east-1.amazonaws.com) specifically for US East (N. Virginia) region.

zhangzhx:
By the way, the Region extraction by the method fromValue() is based on Location Constraint column in the link http://docs.aws.amazon.com/general/latest/gr/rande.html#s3_region.

So unless S3 officially admits us-east-1 as the Location Constraint can we add that.

nterry:
@zhangzhx Its still broken. All other services use us-east-1. The region column in that table mentions 'us-east-1' but makes no mention of 'US', so the fact that you honor 'US' actually already goes against this policy you just mentioned.

nterry:
@zhangzhx I call your attention to the following (in the note above the table):

Amazon S3 renamed the US Standard region to the US East (N. Virginia) region to be consistent with AWS regional naming conventions

There is some verbiage to support my argument.

zhangzhx:
The fromValue() method does honor 'EU'. See

https://github.com/aws/aws-sdk-java/blob/master/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/model/Region.java#L94

There is a lot of confusion behind of the scene. The US Standard was the legacy name for the global endpoint s3.amazonaws.com. They updated to us-east-1 to be consistent with other services as described in the doc.

nterry:
@zhangzhx **Exactly** my point. US_Standard is legacy. This brings it up to modern parlance without breaking the usage of 'US'.

Also, I saw that line after looking further, which is why i removed that part of my comment.

manikandanrs:
I agree that currently there is some confusion as S3 recently renamed one of its endpoints. Although there is a renaming, S3 doesn't return "us-east-1" is any of its API. For instance [Get Bucket Location](http://docs.aws.amazon.com/AmazonS3/latest/API/RESTBucketGETlocation.html) API still returns empty string for buckets in US Standard. Unless you have a problem with round tripping the value "us-east-1" from S3, I don't understand how this change it going to help the customers.

nterry:
@manikandanrs This class is to parse a region into an S3 compatible form. The use case is from transitioning from a service that uses us-east-1 (most of them) to S3 without additional boilerplate code. Using this class won't break the above example you gave, it just adds another option for those used to using us-east-1 as a region which is very common.

manikandanrs:
Agreed @nterry . Let me talk to the team and see how we can make things better here.

nterry:
@manikandanrs Thank you. Let me know what you get.

manikandanrs:
Is passing in "us-east-1" to set the region on AmazonS3Client the concern here ?. If so, would the below code help 

`
AmazonS3Client s3Client = new AmazonS3Client();
client.setRegion(RegionUtils.getRegion("us-east-1"));
`

nterry:
@manikandanrs Thats not the issue. Im not trying to use the region API. Im trying to parse an arbitrary string to the appropriate region. I dont want to perform an API call when I can effectively this method.

manikandanrs:
@nterry I am trying to understand where you want to use the constructed Region enum ? As far as I see, this enum is used only in the output and accepted as input in only CreateBucket which doesn't require location constraint for US Standard.

It would be better if you could give information about where you plan to use the constructed Region enum.

nterry:
@manikandanrs Its used for cross-services purposes within our software. It has nothing to do with the region API. I am taking a user-provided string and trying to parse it into an S3 region, failing and returning an error if it isn't a valid region. I don't want to make an API call if i can avoid it. Currently, I have boilerplate code to handle this edge case, i.e.

`if ("us-east-1".equals(userSuppliedRegion) {
  return Region.US_Standard;
}
return Region.fromValue(userSuppliedRegion);`

I want to be able to remove the top 3 lines from the above pseudo-code

manikandanrs:
@nterry I understand the intention to remove the boiler plate code which is great. But I am failing to understand what you planned to do with the returned Region enum. May be there is a better way to handle your scenario instead of using Region enum.

nterry:
@manikandanrs I'm using it for other API calls, but I fail to see how that relates to the issue at hand. Regardless of the usage of the Region object, the fromValue() method is missing an endpoint string for parsing...

breedloj:
Hey Nicholas, thanks for your follow ups.  We're trying to better understand your use case to see if there's a better solution for what you're trying to do, or if perhaps there's a use case here that we had not previously considered.  The S3 specific region enum is not widely used in the SDK and we have not seen many customers try to snag an S3 region enum for use elsewhere.  We completely understand your confusion regarding our treatment of us-standard and us-east-1 and agree that improvements can be made here, both in the documentation and perhaps in the SDK.  That said if you could share some more details of how you're using the enum that would be very helpful for us when considering any potential changes to the SDK, now and moving forward.

nterry:
@breedloj Im not sure how to explain it any differently..... I am simply wanting to avoid making an API call when normalizing user input... Thats literally it...

This is simply an inconsistency in the API that should have been included from the get-go, regardless of how many users use this particular class. Most of the other services will recognize the string 'us-east-1' and US_Standard is in the same region, even if it goes by another name in your API.

nterry:
@breedloj  @manikandanrs Any updates?

shabbira08:
JJJJ

shorea:
We discussed this yesterday and I believe we are ready to accept this change. @nterry, just so I better understand your use case. What do you do with the S3 Region enum once created?

nterry:
@shorea I just need it to be consistent with other aws apis. I ask the region in a polymorphic way (as I support multiple AWS services) and it is really annoying to make a custom check for S3.

nterry:
@shorea Are you going to merge it?

shorea:
Yep sorry. Let me follow up on the internal CR.

shorea:
Got approval. Merging this now, we have a release on Tuesday if that's alright.

shorea:
@nterry this was released today. Thanks for the contribution!

PR #761, opened 2016-06-28
Add an overload to TransferManager.copy(...) with a 'srcS3' parameter

spew:
When performing a copy of an object across regions the source bucket is in a different
location than the AmazonS3 instance that was used to construct TransferManager. Before
this change when doing such a copy users would see a "301 permanently moved" response
from S3.


shorea:
Merging, this will be included in the next release of the SDK. Thanks for the contribution!

PR #807, opened 2016-08-10
Adding method existsNullOrEmpty

carloslimasis:
This method can be used to verify whether empty or null strings at once. For example:

Using the method isNullOrBlank:
StringUtils.isNullOrBlank (string1) && StringUtils.isNullOrBlank (string2) && StringUtils.isNullOrBlank (string3)

Using the method existsNullOrEmpty:
StringUtils.existsNullOrEmpty (string1, string2, string3)


varunnvs92:
Thanks for your PR. Unfortunately we don't support extra utility methods unless it is requested by many customers. There should be better libraries that have the similar functionality as the method you provided, so there is no need for the AWS SDK for Java to support it. Thank you again.

PR #814, opened 2016-08-16
Fix wrong property filename in log by replace on constant

brahinets:


spfink:
Thanks for the pull request! We are very appreciative! We'll get this merged in quickly.

PR #854, opened 2016-09-23
create parent dir also in parallel downloads, fixes #853

### LINKED ISSUE #853: Discrepancy between TransferManager's parallel and single-part downloads.
For com.amazonaws.services.s3.transfer.TransferManager#download(com.amazonaws.services.s3.model.GetObjectRequest, java.io.File), when the destination directory of a file does not exist, single-part download will create one in the first lines of: com.amazonaws.services.s3.internal.ServiceUtils#downloadToFile

A multi-part parallel download will fail when attempting to lock that file when combining the parts:
```
com.amazonaws.services.s3.transfer.exception.FileLockException: java.io.FileNotFoundException: /tmp/non_existing_subdir/zero (No such file or directory)
	at com.amazonaws.services.s3.internal.FileLocks.lock(FileLocks.java:76)
	at com.amazonaws.services.s3.transfer.DownloadCallable.truncateDestinationFileIfNecessary(DownloadCallable.java:222)
	at com.amazonaws.services.s3.transfer.DownloadCallable.combineFiles(DownloadCallable.java:200)
	at com.amazonaws.services.s3.transfer.DownloadCallable.downloadInParallel(DownloadCallable.java:193)
	at com.amazonaws.services.s3.transfer.DownloadCallable.call(DownloadCallable.java:131)
	at com.amazonaws.services.s3.transfer.DownloadCallable.call(DownloadCallable.java:52)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)
Caused by: java.io.FileNotFoundException: /tmp/non_existing_subdir/zero (No such file or directory)
	at java.io.RandomAccessFile.open0(Native Method)
	at java.io.RandomAccessFile.open(RandomAccessFile.java:316)
	at java.io.RandomAccessFile.<init>(RandomAccessFile.java:243)
	at com.amazonaws.services.s3.internal.FileLocks.lock(FileLocks.java:70)
	... 9 more

```
Either of these behaviours could be considered correct (with the single-part behaviour being the more user-friendly one), but they should stay consistent.
I don't think there's a choice other than implementing the dir creation in parallel download, the other option (removing parent dir creation) will certainly break user workflows.

pbruski:


aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/transfer/DownloadCallable.java
```diff
@@ -197,7 +196,15 @@ private void downloadInParallel(int partCount) throws Exception {
      * Merges all the individual part Files into dstFile
      */
     private void combineFiles() throws Exception {
-        truncateDestinationFileIfNecessary();
+        final File parentDirectory = dstfile.getParentFile();
+        if (parentDirectory == null || parentDirectory.exists()) {
+            truncateDestinationFileIfNecessary();
+        } else {
+            if (!parentDirectory.mkdirs()) {
+                throw new AmazonClientException("Unable to create directory in the path" + parentDirectory.getAbsolutePath());
```
drainx: Space or ": " needed in exception message

aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/transfer/DownloadCallable.java
```diff
@@ -197,7 +196,15 @@ private void downloadInParallel(int partCount) throws Exception {
      * Merges all the individual part Files into dstFile
      */
     private void combineFiles() throws Exception {
-        truncateDestinationFileIfNecessary();
+        final File parentDirectory = dstfile.getParentFile();
+        if (parentDirectory == null || parentDirectory.exists()) {
+            truncateDestinationFileIfNecessary();
+        } else {
+            if (!parentDirectory.mkdirs()) {
```
drainx: Make this just an else if

aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/transfer/DownloadCallable.java
```diff
@@ -217,7 +224,7 @@ private void combineFiles() throws Exception {
      * operation), adjust the file length so that the part starts writing from
      * the correct position.
      */
-    private void truncateDestinationFileIfNecessary() throws IOException {
+    private void truncateDestinationFileIfNecessary() {
```
drainx: Why remove the IOException and force up into the general exception?

pbruski: IOException was not thrown from this method. It only thrown unchecked exception.

aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/transfer/DownloadCallable.java
pbruski: Heh, I copied this from ServiceUtils. I'll fix the other place too.

aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/transfer/DownloadCallable.java
pbruski: good catch.

kiiadi:
Minor optimization but it'd be nice to have the directory creation for both in a single util method so we're consistent with it's handling. Perhaps a `createParentDirectory` method that both single and multipart downloads can use.

kiiadi:
Actually looking at this again, I don't think we need this. In the most recent version of the SDK we made an update to the multipart download so that file parts are downloaded in the same directory as the target file (ie: dstFile) in this case the parent should already exist. `ServiceUtils.downloadToFile` gets called in both `DownloadCallable` and `DownloadPartCallable`.

If you upgrade to 1.11.37 you should not need this explicit create for multi-part files.

pbruski:
Hmm, I'd swear I've tested it with the most recent version too and it failed. I'll have a look at it again after the weekend.

kiiadi:
1.11.37 was only released yesterday.

pbruski:
No, 37 is still broken. What you say about the new behaviour is all true, except that the exception is thrown from the locking code, so potentially before the parent dir gets asynchonously created in the download callables.

kiiadi:
That makes sense, I guess we could have solved it by just pushing the call to `truncateDestinationFileIfNecessary()` down into the loop around the Futures - but I like the explicit approach you've gone with.

Looks good to me - this fix should go out with our next release.

aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/transfer/DownloadCallable.java
```diff
@@ -197,7 +196,10 @@ private void downloadInParallel(int partCount) throws Exception {
      * Merges all the individual part Files into dstFile
      */
     private void combineFiles() throws Exception {
-        truncateDestinationFileIfNecessary();
+        if (!ServiceUtils.createParentDirectoryIfNecessary(dstfile)) {
```
zymzxq: This won't fix the problem. `futureFiles` starts download concurrently, which may happen before this is excuted

kiiadi: The future file get also creates the parent directory if necessary I believe. So which ever one starts first should create the parent directory and hopefully resolve the issue.

pbruski: Unfortunately, the temp file is created before the ServiceUtils parent directory creation is done. So, yeah, the fix is not complete.

PR #855, opened 2016-09-23
fixed project description in pom.xml

cmello:


kiiadi:
Thanks for the correction.

PR #858, opened 2016-09-27
prevent LambdaSerializationException on toString()

max-preuschen:
When calling `toString()` on interface class (after `LambdaInvokerFactory.build`), exception gets thrown: `No LambdaFunction annotation for method toString`.
=> `toString` should be object method, **not invoked** method

**e.g.**
```Java
interface MyService {
	@LambdaFunction(functionName = "lambdaFunction")
	void lambdaFunction();
}
```
```Java
public static void main(String[] args) {
    AWSLambdaClient lambdaClient = new AWSLambdaClient();
    lambdaClient.configureRegion(Regions.US_EAST_1);
    MyService myService = LambdaInvokerFactory.build(MyService.class, lambdaClient);

    System.out.println(myService);
}

```
## Exception:
```
Exception in thread "main" com.amazonaws.services.lambda.invoke.LambdaSerializationException: No LambdaFunction annotation for method toString
	at com.amazonaws.services.lambda.invoke.LambdaInvokerFactory$LambdaInvocationHandler.validateInterfaceMethod(LambdaInvokerFactory.java:134)
	at com.amazonaws.services.lambda.invoke.LambdaInvokerFactory$LambdaInvocationHandler.invoke(LambdaInvokerFactory.java:117)
	at $Proxy9.toString(Unknown Source)
```

### Annotating doesn't work either:
```Java
interface MyService
{
	@LambdaFunction(functionName = "lambdaFunction")
	void lambdaFunction();

	@LambdaFunction(functionName = "toString")
	String toString();
}
```

### DIFF
--- a/aws-java-sdk-lambda/src/main/java/com/amazonaws/services/lambda/invoke/LambdaInvokerFactory.java
+++ b/aws-java-sdk-lambda/src/main/java/com/amazonaws/services/lambda/invoke/LambdaInvokerFactory.java
@@ -113,14 +113,13 @@ public final class LambdaInvokerFactory {
 
         @Override
         public Object invoke(Object proxy, Method method, Object[] args) throws Throwable {
-
-            LambdaFunction annotation = validateInterfaceMethod(method, args);
-
-            InvokeRequest invokeRequest = buildInvokeRequest(method, annotation, args == null ? null : args[0]);
-
-            InvokeResult invokeResult = awsLambda.invoke(invokeRequest);
-
-            return processInvokeResult(method, invokeResult);
+            if(method.getName().equals("toString")) return this.toString();
+            else {
+                LambdaFunction annotation = validateInterfaceMethod(method, args);
+                InvokeRequest invokeRequest = buildInvokeRequest(method, annotation, args == null ? null : args[0]);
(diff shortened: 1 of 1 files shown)


shorea:
Looks good to me. Just curious what the use case is.

max-preuschen:
I encountered this exception a lot while debugging (the debugger automatically calls `toString()` to visualise instances.

I tried to use the library in Scala and creation of the object caused the `LambdaSerializationException` as well (it seems `toString()` is called implicitly somewhere...?)
```Scala
val myService = LambdaInvokerFactory.build(classOf[MyService], lambdaClient) //throws exception
```

Getting this exception was very confusing and took me quite some time to figure out what was going on, since I didn't call `toString()` explicitly. 
Somehow, it just seems wrong to get an exception, if you call `toString()` on any object.

Would be cool, if it could make it into the next release 👍 

The AWS libraries are really great and make development so much easier!

shorea:
Sounds good, thanks for the explanation. I've got to get this approved internally but hopefully it can make it out today, if not it will be released on Thursday.

shorea:
Merging this in. Thanks for the contribution!

PR #859, opened 2016-09-28
Moving parent directory creation up

kiiadi:
Fix for #853 so that parent directory is always created before any files (tmp or otherwise) are created during parallel file downloads.

Follow-on from #854

### DIFF
--- a/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/transfer/DownloadCallable.java
+++ b/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/transfer/DownloadCallable.java
@@ -125,6 +125,7 @@ final class DownloadCallable implements Callable<File> {
             }
 
             download.setState(TransferState.InProgress);
+            ServiceUtils.createParentDirectoryIfNecessary(dstfile);
 
             if (isDownloadParallel) {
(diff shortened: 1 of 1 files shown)


aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/transfer/DownloadCallable.java
```diff
@@ -174,6 +174,8 @@ private void downloadInParallel(int partCount) throws Exception {
             lastFullyMergedPartNumber = 0;
         }
 
+        ServiceUtils.createParentDirectoryIfNecessary(dstfile);
```
pbruski: Why not move it all the way up to line 128? We really don't want to run into these inconsistencies again ;-)

kiiadi: My initial thought was try to do this as late as possible - moving it up to line 128 doesn't actually buy us anything in this context because it'll get created down both paths (single/multi) anyway.

However - I can't think of a good reason *not* to move it up to earlier in the piece and it might even be useful from a permissions perspective - if the app doesn't have permissions to write might as well bomb before we've even started trying to download.

PR #868, opened 2016-10-04
update query request for expression names, values

amcp:
@hansonchar @manikandanrs I tried setting expression names and values directly on QuerySpec and it didnt work. Probably, QueryExpressionSpec works, but this should be fixed anyway. It can impede people from using Key Condition expressions.


amcp:
withExpressionSpec calls withNameMap and withValueMap, so key condition expression query in document sdk was always broken...

Issue #871, opened 2016-10-05
Race condition during parent directory creation in S3 download


pbruski:
When ServiceUtils.createParentDirectoryIfNecessary is called concurrently by several threads. File.mkdirs() call may fail because the directory already exists, although it hasn't existed before the mkdirs() call.
This no longer causes problems in the latest versions of SDK, because we create the parent dir upfront, but needs fixing nevertheless.

kiiadi:
Fixed by fix to [ServiceUtils.createParentDirectory](https://github.com/aws/aws-sdk-java/blob/master/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/internal/ServiceUtils.java#L343) closed PR #872

PR #872, opened 2016-10-05
#871 fix race condition during directory creation and remove unnecessary log

pbruski:


aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/transfer/DownloadCallable.java
```diff
@@ -203,13 +203,6 @@ private void combineFiles() throws Exception {
             File partFile = f.get();
             ServiceUtils.appendFile(partFile, dstfile);
             download.updatePersistableTransfer(++lastFullyMergedPartNumber);
-            try {
```
pbruski: appendFile already deletes the part file

aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/internal/ServiceUtils.java
```diff
@@ -346,7 +346,7 @@ public static boolean createParentDirectoryIfNecessary(final File file) {
         if (parentDirectory == null || parentDirectory.exists()) {
             return false;
         }
-        if (parentDirectory.mkdirs()) {
+        if (parentDirectory.mkdirs() || parentDirectory.isDirectory()) {
             return true;
```
kiiadi: The return-type here is never actually used and doesn't really add any value; I propose we change the signature of this method to return void instead: 

```java
public static void createParentDirectoryIfNecessary(final File file) {
    final File parentDirectory = file.getParentFile();
    if (parentDirectory == null || parentDirectory.exists() || parentDirectory.mkdirs()) {
        return;
    }

    if (!parentDirectory.exists()) {
        throw new AmazonClientException("Unable to create directory in the path: " + parentDirectory.getAbsolutePath());
    }
}
```

kiiadi:
Pushed a fix to [ServiceUtils.createParentDirectory](https://github.com/aws/aws-sdk-java/blob/master/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/internal/ServiceUtils.java#L343).

Issue #874, opened 2016-10-07
Overflow problem in ExponentialBackoffStrategy


yairzaslavsky:
return (originalRequest, exception, retriesAttempted) -> (retriesAttempted > maxRetries) ? maxBackoffTime :
                Math.min(((1 << retriesAttempted) \* baseDelay), maxBackoffTime);

This might cause an overflow (negative integer value due to the overflow) and to return a negative delay value.

dagnir:
This is fixed in version [`1.11.49`](https://aws.amazon.com/releasenotes/2981276408528428). Thanks again for your help in finding and fixing the issue!

PR #875, opened 2016-10-07
Fix overflow problem (#874)

yairzaslavsky:

My fix simply changes this to (1L << retriesAttempted) , which makes
the multiplication an a long value, avoiding the overflow

### DIFF
--- a/aws-java-sdk-core/src/main/java/com/amazonaws/retry/PredefinedBackoffStrategies.java
+++ b/aws-java-sdk-core/src/main/java/com/amazonaws/retry/PredefinedBackoffStrategies.java
@@ -103,7 +103,7 @@ public class PredefinedBackoffStrategies {
                                          AmazonClientException exception,
                                          int retriesAttempted) {
             return (retriesAttempted > MAX_RETRIES) ? maxBackoffTime :
-                    Math.min(((1 << retriesAttempted) * baseDelay), maxBackoffTime);
+                    Math.min(((1L << retriesAttempted) * baseDelay), maxBackoffTime);
         }
     }
 
--- a/aws-java-sdk-core/src/main/java/com/amazonaws/retry/PredefinedBackoffStrategies.java
+++ b/aws-java-sdk-core/src/main/java/com/amazonaws/retry/PredefinedBackoffStrategies.java
@@ -43,6 +43,13 @@ public class PredefinedBackoffStrategies {
     /** Maximum retry limit.  Avoids integer overflow issues. **/
     private static final int MAX_RETRIES = 30;
 
+    private static final long exponentialDelayHelper(int retriesAttempted, int baseDelay, int maxBackoffTime) {
+      long potentialWait = (1L << retriesAttempted) * baseDelay;
+      return (retriesAttempted > MAX_RETRIES) ? maxBackoffTime :
+              (potentialWait < 0  ? maxBackoffTime :
+              Math.min(potentialWait, maxBackoffTime));
+    }
+
     public static class FullJitterBackoffStrategy implements RetryPolicy.BackoffStrategy {
 
         private final int baseDelay;
@@ -59,9 +66,8 @@ public class PredefinedBackoffStrategies {
         public long delayBeforeNextRetry(AmazonWebServiceRequest originalRequest,
                                          AmazonClientException exception,
                                          int retriesAttempted) {
-            int ceil = (retriesAttempted > MAX_RETRIES) ? maxBackoffTime :
-                    Math.min(baseDelay * (1 << retriesAttempted), maxBackoffTime);
-            return random.nextInt(ceil + 1);
+            long ceil =  exponentialDelayHelper(retriesAttempted, baseDelay, maxBackoffTime);
+            return random.nextLong() % (ceil + 1);
         }
     }
 
@@ -81,9 +87,8 @@ public class PredefinedBackoffStrategies {
         public long delayBeforeNextRetry(AmazonWebServiceRequest originalRequest,
                                         AmazonClientException exception,
                                         int retriesAttempted) {
-            int ceil = (retriesAttempted > MAX_RETRIES) ? maxBackoffTime
-                    : Math.min(maxBackoffTime, baseDelay * (1 << retriesAttempted));
-            return (ceil / 2) + random.nextInt((ceil / 2) + 1);
+            long ceil =  exponentialDelayHelper(retriesAttempted, baseDelay, maxBackoffTime);
+            return (ceil / 2) + (random.nextLong() % ((ceil / 2) + 1));
         }
     }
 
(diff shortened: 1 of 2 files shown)


aws-java-sdk-core/src/main/java/com/amazonaws/retry/PredefinedBackoffStrategies.java
```diff
@@ -103,7 +103,7 @@ public long delayBeforeNextRetry(AmazonWebServiceRequest originalRequest,
                                          AmazonClientException exception,
                                          int retriesAttempted) {
             return (retriesAttempted > MAX_RETRIES) ? maxBackoffTime :
-                    Math.min(((1 << retriesAttempted) * baseDelay), maxBackoffTime);
+                    Math.min(((1L << retriesAttempted) * baseDelay), maxBackoffTime);
```
kiiadi: Thanks for the PR - good catch. I think in terms of implementation though we're potentially just dodging the problem - as soon as you get MAX_RETRIES >= 63 we're going to have the same issue (ie: Long's overflow) - we might be better adding another call to Math to help us. This should resolve the issue:

`Math.max(1, 1 << retriesAttempted)`

Also we should use the same with all the pre-canned back-off strategies.

kiiadi: Alternatively we could do this which might actually be better:

```java
int exponential = 1 << retriesAttempted;
return exponential >= 0 ? exponential : Integer.MAX_VALUE;
```

yairzaslavsky: @kiiadi Thanks for your comments,
On 1.11.40 -  MAX_RERTIES is set to 30 , so I did not handle the potential overflow of long, as it does not reach the value of 63.

Using the exponential value  as an int and multiply it by baseDelay might still cause an Integer overflow.

Since the method dleayBeforeNextRetry returns long, I think it's better to resort to longs as soon as possible in the method's code.

I agree that other strategies should be fixed as well.

aws-java-sdk-core/src/main/java/com/amazonaws/retry/PredefinedBackoffStrategies.java
```diff
@@ -77,7 +77,7 @@ public EqualJitterBackoffStrategy(final int baseDelay,
             this.maxBackoffTime = ValidationUtils.assertIsPositive(maxBackoffTime, "Max backoff");
         }
 
-        @Override
+        @Override*5
```
kiiadi: Is this a typo?

aws-java-sdk-core/src/main/java/com/amazonaws/retry/PredefinedBackoffStrategies.java
```diff
@@ -102,8 +102,10 @@ public ExponentialBackoffStrategy(final int baseDelay,
         public long delayBeforeNextRetry(AmazonWebServiceRequest originalRequest,
                                          AmazonClientException exception,
                                          int retriesAttempted) {
+            long potentialWait = (1L << retriesAttempted) * baseDelay;
             return (retriesAttempted > MAX_RETRIES) ? maxBackoffTime :
-                    Math.min(((1 << retriesAttempted) * baseDelay), maxBackoffTime);
+                    (potentialWait < 0  ? maxBackoffTime :
```
kiiadi: Reckon we can push this into a static private method that the various strategies could use?

aws-java-sdk-core/src/main/java/com/amazonaws/retry/PredefinedBackoffStrategies.java
yairzaslavsky: Yes.
Already suggested a fix.

aws-java-sdk-osgi/dependency-reduced-pom.xml
```diff
@@ -3,7 +3,7 @@
   <parent>
     <artifactId>aws-java-sdk-pom</artifactId>
     <groupId>com.amazonaws</groupId>
-    <version>1.11.41</version>
+    <version>1.11.42-SNAPSHOT</version>
```
kiiadi: Did you mean to commit changes to the pom file here? They seem unrelated.

aws-java-sdk-core/src/main/java/com/amazonaws/retry/PredefinedBackoffStrategies.java
```diff
@@ -43,6 +43,13 @@
     /** Maximum retry limit.  Avoids integer overflow issues. **/
     private static final int MAX_RETRIES = 30;
 
+    private static final long exponentialDelayHelper(int retriesAttempted, int baseDelay, int maxBackoffTime) {
+      long potentialWait = (1L << retriesAttempted) * baseDelay;
+      return (retriesAttempted > MAX_RETRIES) ? maxBackoffTime :
+              (potentialWait < 0  ? maxBackoffTime :
+              Math.min(potentialWait, maxBackoffTime));
+    }
+
```
kiiadi: /Nitpick: do you mind if we push this method to the bottom of the class?

Also I suspect we can get away with just taking in "retriesAttempted" as a parameter since it's the `1 << retriesAttempted` line that we're trying to wrap. What if we did something like this:

```java
private static int calculateExponentialMultiplier(int retriesAttempted) {
    int multiplier = 1 << retriesAttempted;
    return multiplier >= 0 ? multiplier : Integer.MAX_VALUE;
}
```

aws-java-sdk-osgi/dependency-reduced-pom.xml
yairzaslavsky: My bad about the pom, thanks

aws-java-sdk-core/src/main/java/com/amazonaws/retry/PredefinedBackoffStrategies.java
yairzaslavsky: I would prefer to stick to long here,  if you take the multiplier and multiply it by a baseDelay you will still have an overflow.
I would prefer to handle the overflow inside the helper method.

What do you think about this?

yairzaslavsky: Submitted a patch with no pom.xml change and moving the static method down in the file.

aws-java-sdk-core/src/main/java/com/amazonaws/retry/PredefinedBackoffStrategies.java
```diff
@@ -147,4 +144,10 @@ public final long delayBeforeNextRetry(AmazonWebServiceRequest originalRequest,
         }
     }
 
-}
\ No newline at end of file
+    private static final long exponentialDelayHelper(int retriesAttempted, int baseDelay, int maxBackoffTime) {
```
kiiadi: /nitpick - can we rename this method so that it's a verb? Something like "calculateExponentialDelay"

yairzaslavsky: Looks like that as a part of running "mvn" locally the  pom.xml file gets changed.
I will rebase again against master and get rid of this change.

I am fine with the name suggestion.

PR #880, opened 2016-10-10
Fix overflow problem (#874)

yairzaslavsky:

My fix checks if the multiplication yields an overflow, and if so will return
the maxBackoffTime, if not it will return the minimum between the multiplication
and the max backoff time


dagnir:
Hey @yairzaslavsky, can you update your original `issue_874_fix` branch instead? It will make reviewing your changes easier. Thank you!

dagnir:
@yairzaslavsky Thanks!

(closed, dup of https://github.com/aws/aws-sdk-java/pull/875)

PR #881, opened 2016-10-11
Fix overflow problem (#874)

yairzaslavsky:

Comment for maintainers:
There was too much noise on my previous pull-request, I apologize for that, 
I submitted a pull request with your suggestions.

### DIFF
--- a/aws-java-sdk-core/src/main/java/com/amazonaws/retry/PredefinedBackoffStrategies.java
+++ b/aws-java-sdk-core/src/main/java/com/amazonaws/retry/PredefinedBackoffStrategies.java
@@ -14,12 +14,12 @@
  */
 package com.amazonaws.retry;
 
+import java.util.Random;
+
 import com.amazonaws.AmazonClientException;
 import com.amazonaws.AmazonServiceException;
 import com.amazonaws.AmazonWebServiceRequest;
 import com.amazonaws.util.ValidationUtils;
-
-import java.util.Random;
 
 /**
  * This class includes a set of pre-defined backoff policies.
@@ -59,8 +59,8 @@ public class PredefinedBackoffStrategies {
         public long delayBeforeNextRetry(AmazonWebServiceRequest originalRequest,
                                          AmazonClientException exception,
                                          int retriesAttempted) {
-            long ceil =  calculateExponentialDelay(retriesAttempted, baseDelay, maxBackoffTime);
-            return random.nextLong() % (ceil + 1);
+            int ceil =  (int)calculateExponentialDelay(retriesAttempted, baseDelay, maxBackoffTime);
+            return random.nextInt(ceil + 1);
         }
     }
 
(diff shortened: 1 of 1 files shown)


kiiadi:
This has actually broken the tests - can you take a look?
https://travis-ci.org/aws/aws-sdk-java/jobs/166628623

Suspect it might be the change in logic from `random.nextInt(bound)` to `random.nextLong() % bound`

yairzaslavsky:
@kiiadi  fixed, 

https://travis-ci.org/aws/aws-sdk-java/jobs/166906300  fails due to permGen issue.

aws-java-sdk-core/src/main/java/com/amazonaws/retry/PredefinedBackoffStrategies.java
```diff
@@ -59,8 +59,7 @@ public FullJitterBackoffStrategy(final int baseDelay,
         public long delayBeforeNextRetry(AmazonWebServiceRequest originalRequest,
                                          AmazonClientException exception,
                                          int retriesAttempted) {
-            int ceil = (retriesAttempted > MAX_RETRIES) ? maxBackoffTime :
-                    Math.min(baseDelay * (1 << retriesAttempted), maxBackoffTime);
+            int ceil =  (int)calculateExponentialDelay(retriesAttempted, baseDelay, maxBackoffTime);
```
kiiadi: I'm not a huge fan of the casting between long and int - if we go over int's bound we're going to run into the same overflow problem when we do this cast.

aws-java-sdk-core/src/main/java/com/amazonaws/retry/PredefinedBackoffStrategies.java
```diff
@@ -81,9 +80,8 @@ public EqualJitterBackoffStrategy(final int baseDelay,
         public long delayBeforeNextRetry(AmazonWebServiceRequest originalRequest,
                                         AmazonClientException exception,
                                         int retriesAttempted) {
-            int ceil = (retriesAttempted > MAX_RETRIES) ? maxBackoffTime
-                    : Math.min(maxBackoffTime, baseDelay * (1 << retriesAttempted));
-            return (ceil / 2) + random.nextInt((ceil / 2) + 1);
+            int ceil =  (int)calculateExponentialDelay(retriesAttempted, baseDelay, maxBackoffTime);
+            return (ceil / 2) + (long)random.nextInt((ceil / 2) + 1);
```
kiiadi: Do we need this cast to (long) ?

kiiadi:
One other thing - I'd like to try to reproduce the overflow scenario you were seeing in a test case so that we're sure this has solved it and we don't regress in the future. Could you add a couple of test cases?

Thanks so much for the time spent on this - it'll be good to have this issue resolved.

yairzaslavsky:
I was away for  some time and could not work on that, I decided to use a different approach. I will push a different approach + unit test.

varunnvs92:
@yairzaslavsky 
We actually implemented a different approach internally as other customers were waiting for the fix. The fix will be included in tomorrow's release. Thank you for your effort and contribution.

kiiadi:
It's worth pointing out that the approach we ended up going with was very similar to your most recent PR - so thanks for helping us find and fix this issue!

PR #889, opened 2016-10-14
Tighter control of when InstanceProfileCredentialsProvider fetches

### LINKED ISSUE #884: InstanceProfileCredentialsProvider refresh is by default off
Both [DefaultAWSCredentialsProviderChain](https://github.com/aws/aws-sdk-java/blob/master/aws-java-sdk-core/src/main/java/com/amazonaws/auth/DefaultAWSCredentialsProviderChain.java#L48) and the static instance of [InstanceProfileCredentialsProvider](https://github.com/aws/aws-sdk-java/blob/master/aws-java-sdk-core/src/main/java/com/amazonaws/auth/InstanceProfileCredentialsProvider.java#L52) use the no arg constructor which does not periodically refresh the credentials on an EC2. I can't think of a time when I would want my EC2 credentials to expire. Could the default be made to refresh or provide a way to set/update this?

On an EC2, under heavy load, after about an hour I am getting:

```
com.amazonaws.AmazonClientException: Unable to load AWS credentials from any provider in the chain
at com.amazonaws.auth.AWSCredentialsProviderChain.getCredentials(AWSCredentialsProviderChain.java:131)
at our.code:0)
```

kiiadi:
Fixes #884

### DIFF
--- a/aws-java-sdk-core/src/main/java/com/amazonaws/auth/InstanceProfileCredentialsProvider.java
+++ b/aws-java-sdk-core/src/main/java/com/amazonaws/auth/InstanceProfileCredentialsProvider.java
@@ -60,6 +60,8 @@ public class InstanceProfileCredentialsProvider implements AWSCredentialsProvide
      */
     private volatile ScheduledExecutorService executor;
 
+    private volatile boolean getCredentialsCalled = false;
+
     /**
      * @deprecated for the singleton method {@link #getInstance()}.
      */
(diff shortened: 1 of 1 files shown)


aws-java-sdk-core/src/main/java/com/amazonaws/auth/InstanceProfileCredentialsProvider.java
```diff
@@ -76,7 +78,22 @@ public InstanceProfileCredentialsProvider() {
...
+     *
+     * @param refreshCredentialsAsync
+     *            true if credentials needs to be refreshed asynchronously else
+     *            false.
+     * @param eagerlyRefreshCredentialsAsync
+     *            when set to false will not attempt to refresh credentials asynchronously
+     *            until after a call has been made to {@link #getCredentials()} - ensures that
+     *            {@link EC2CredentialsFetcher#getCredentials()} is only hit when this CredentialProvider is actually required
+     */
+    public InstanceProfileCredentialsProvider(boolean refreshCredentialsAsync, final boolean eagerlyRefreshCredentialsAsync) {
```
JeffAtDeere: This doesn't do what you think it does. By the time `getCredentialsCalled` is set to true in `getCredentials()`, this constructor has long since been run and executed thus won't execute again to pick up the change in `getCredentialsCalled`. It's effectively dead code.

aws-java-sdk-core/src/main/java/com/amazonaws/auth/InstanceProfileCredentialsProvider.java
```diff
@@ -85,7 +102,7 @@ public InstanceProfileCredentialsProvider(boolean refreshCredentialsAsync) {
                 @Override
                 public void run() {
                     try {
-                        credentialsFetcher.getCredentials();
+                        if (eagerlyRefreshCredentialsAsync || getCredentialsCalled) credentialsFetcher.getCredentials();
```
JeffAtDeere: I don't think this does what is intended. The goal is to be safe (as in, not fail every invocation of `run()` when not on an EC2). Given `new InstanceProfileCredentialsProvider(true)` or `new InstanceProfileCredentialsProvider(true, true)` does not do that because it will attempt to refresh. The only safe option for EC2 and non-EC2 is `new InstanceProfileCredentialsProvider(true, false)` so that should really be the only option (i.e. remove `eagerlyRefreshCredentialsAsync`).

aws-java-sdk-core/src/main/java/com/amazonaws/auth/InstanceProfileCredentialsProvider.java
```diff
@@ -121,6 +138,7 @@ protected void finalize() throws Throwable {
 
     @Override
     public AWSCredentials getCredentials() {
+        getCredentialsCalled = true;
         return credentialsFetcher.getCredentials();
     }
```
JeffAtDeere: Is it sufficient that `getCredentials()` was called? A stronger statement would be '`getCredentials()` was called and succeeded' such as
```java
     public AWSCredentials getCredentials() {
          AWSCredentials  creds = credentialsFetcher.getCredentials();
          getCredentialsCalled = true;
          return creds;
      }
```

kiiadi:
This has been released see here https://github.com/aws/aws-sdk-java/blob/master/aws-java-sdk-core/src/main/java/com/amazonaws/auth/InstanceProfileCredentialsProvider.java#L91

Issue #902, opened 2016-10-25
Orphaned log message “[Fatal Error] :1:1: Premature end of file.”


ssurabhi:
We are occasionally seeing the above message in stdout log file. There are 2 issues with the message
1. There is no context in the message, neither stack trace nor the name of the xml file that it failed to parse. 
2. That message is logged to std out file rather a  file configured thru application logger. 

After debugging, we found that it is caused due to parsing an empty xml type response from one  (SNS) of the aws services. Upon digging into logs we found that application is receiving an empty error response when http status is 503. But the code is still going ahead and parsing the empty error response resulting in error “premature end of file”. This is triggered from com.amazonaws.util.XpathUtils.documentFrom(InputSource). When an empty response is run with this code, you would see the output is going to std out, which is written from com.sun.org.apache.xerces.internal.util.DefaultErrorHandler.printError().

Here is the code snippet that we can use to reproduce this.   

`System.setProperty("javax.xml.parsers.DocumentBuilderFactory", "com.sun.org.apache.xerces.internal.jaxp.DocumentBuilderFactoryImpl"); XpathUtils.documentFrom("");`

Proposed Change: We want the message output to application log file with some context around what xml file it has failed to parse with stack trace. 

For example pass custom ErrorHandler that logs message with context using logger. 

```
DocumentBuilderFactory factory = DocumentBuilderFactory.newInstance(); 
DocumentBuilder documentBuilder = factory.newDocumentBuilder(); 
String xml = "test.xml"; 
documentBuilder.setErrorHandler(new ErrorHandler() {     
@Override     
public void warning(SAXParseException exception) throws SAXException {        
 logger.warn(" Error parsing {}", xml, exception);     
}      
@Override     
public void error(SAXParseException exception) throws SAXException {         
logger.error(“ Error parsing {}", xml, exception);     
}      
@Override     
public void fatalError(SAXParseException exception) throws SAXException {         
logger.fatal(“ Error parsing {}", xml, exception);    
 }
 });
```

kiiadi:
Can you post the full log message? Looking at the code when we attempt to parse the response XML we print "Unable to parse HTTP response content" along with the error message [here](https://github.com/aws/aws-sdk-java/blob/master/aws-java-sdk-core/src/main/java/com/amazonaws/http/DefaultErrorResponseHandler.java#L104).

I want to make sure we're in the right place, but this is probably where we want to add the log message - since we're logging anyway.

ssurabhi:
[...]

PR #909, opened 2016-10-28
More verbose logging of xml parse errors to address #902

kiiadi:


shorea:
LGTM. There's a client side header called amz-sdk-invocation-id that uniquely identifies a request. May be useful since request id is not available.

thomashunziker:
I understand that you provide with this fix more visibility. Which is great. However I do not understand that you try to parse an empty string anyway. Would it not make more sense to check if the string is empty to avoid the "Premature end of file" print out to the standard output?

PR #914, opened 2016-11-07
small fixes for reference docs

EronHennessey:
These are a few text fixes for existing reference topics...


dagnir:
Thanks Eron. The `TransferManager` manager changes look good. The `DynamoDB` classes are actually auto generated so I will put in the necessary fix in our code generator.

EronHennessey:
Thanks @dagnir!

dagnir:
The fixes will be in our next release. Thank you!

PR #941, opened 2016-12-12
add profile owasp which enables the dependency-check-maven plugin…

muellerc:
… to check all dependencies for known security vulnerabilities


shorea:
We don't use XmlMapper in the Java SDK so we are unaffected by the vulnerability. +1 for the dependency check though.

muellerc:
Hi Andrew,
thanks for the quick feedback. I will add the suppression of this false-positive and send a new pull request.

PR #942, opened 2016-12-13
core: read the inputstream before the synchronized block

pepribas:
Creating the event reader seems to try to read some bytes from the
passed input stream by the looks of it, in order to determine the
encoding. A read on the http input stream causes network access
and suddenly all threads are blocked on that first synchronized block.
The hack here is reading the whole input stream before the synchronized block.

### DIFF
--- a/aws-java-sdk-core/src/main/java/com/amazonaws/http/StaxResponseHandler.java
+++ b/aws-java-sdk-core/src/main/java/com/amazonaws/http/StaxResponseHandler.java
@@ -84,6 +84,12 @@ public class StaxResponseHandler<T> implements HttpResponseHandler<AmazonWebServ
         InputStream content = response.getContent();
         if (content == null) {
             content = new ByteArrayInputStream("<eof/>".getBytes(StringUtils.UTF8));
+        } else {
+            // Creating the event reader seems to try to read some bytes from the passed input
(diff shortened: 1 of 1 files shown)


shorea:
Worried about the performance implications of this, especially for services like EC2 that can return large XML responses. Perhaps we can store the XMLInputFactory in a ThreadLocal and remove the synchronization altogether. And maybe we can call newInstance once and then clone instead of having to go through the expensive lookup process again (not sure if that's viable, haven't tested it).

pepribas:
@shorea ok I understand.
How do you want me to proceed in order to get rid of the synchronized block?
Should I open an issue referencing this PR?
Does it has sense for me to look into the alternative solutions you have listed?

shorea:
If you're up for looking into the other solutions then feel free to edit this PR in place. Otherwise opening an issue referencing this sounds good to me.

millems:
Sorry about the long delay on this one. Is this still something worth keeping open, or can we just defer to the linked feature request?

pepribas:
I'll close the PR

millems:
Thanks! Please also feel free to try out 2.0 of the SDK to see if you notice any meaningful performance improvement: https://github.com/aws/aws-sdk-java-v2

PR #946, opened 2016-12-14
add profile owasp which enables the dependency-check-maven plugin to …

muellerc:
…check all dependencies for known security vulnerabilities


millems:
Sorry about the long delay on this one. Thanks for your contribution! 

This is something we're looking to integrate into 2.0.x. 1.11.x is dependency frozen right now, so this isn't something we'll be able to integrate right now. We'll use this as a reference when we update 2.0.x, though!

Issue #409, opened 2015-04-15
AmazonSQSAsyncClient does not have constructor taking a RequestMetricCollector


robertclancy:
AmazonSQSAsyncClient does expose the constructors of its superclass that [take a RequestMetricCollector](https://github.com/aws/aws-sdk-java/blob/0adb8ab9a4e73350622ff4938820b8dd2c26dc07/aws-java-sdk-sqs/src/main/java/com/amazonaws/services/sqs/AmazonSQSClient.java#L245). This means any metric collection must be configured at the global level with [AwsSdkMetrics::setMetricCollector](https://github.com/aws/aws-sdk-java/blob/0adb8ab9a4e73350622ff4938820b8dd2c26dc07/aws-java-sdk-core/src/main/java/com/amazonaws/metrics/AwsSdkMetrics.java#L455).

Is this due to concerns about the thread safety of implementations of RequestMetricCollector?

muellerc:
https://github.com/aws/aws-sdk-java/pull/957 should resolve the feature request.

shorea:
Client builders support this. Closing this issue. 

https://github.com/aws/aws-sdk-java/blob/master/aws-java-sdk-sqs/src/main/java/com/amazonaws/services/sqs/AmazonSQSAsyncClientBuilder.java

PR #957, opened 2016-12-22
#409: AmazonSQSAsyncClient does not have constructor taking a Request…

muellerc:
…MetricCollector


shorea:
Can you take a look at https://github.com/aws/aws-sdk-java/blob/master/aws-java-sdk-sqs/src/main/java/com/amazonaws/services/sqs/AmazonSQSAsyncClientBuilder.java. We are trying to encourage use of the builders over the constructors.

muellerc:
Ok, will do.
Should I revert the constructor changes or should I only add the builder changes?

shorea:
The builder already supports it.

muellerc:
Ok, fine. Shouldn't we close the issue in this case? Or is there still something to do which I miss?

shorea:
Sure, sounds good to me. If the builders don't work out for your use case feel free to open another issue.

jpaterso:
The builder pattern doesn't allow overriding the client's behaviour around shutdown, specifically waiting for the the executor service to terminate. Is there any way around this?

PR #966, opened 2016-12-30
314: Binary SDK build does not contain OSGi metadata in Manifest

muellerc:
This pull request will fix issue 314 and add the required OSGI metadata to the individual modules.


millems:
Sorry about the long delay on this one. We're being a bit more strict about package naming in V2 and we'll make a note to make sure we don't repeat this same mistake in V2.

Because many of these actions are code generated, it will be a bit of work to integrate this request into 1.11.x. Are you blocked by this in any major way, or is it okay if we just make sure it's fixed in V2?

muellerc:
It is okay if we just make sure it's fixed in V2.

Best,
Christian

PR #968, opened 2016-12-31
Allow shared pool connections

kdheerendra:
This required for fixing issue of IllegalState exception in shared connections
Issue
http://stackoverflow.com/questions/37303954/com-amazonaws-amazonclientexception-unable-to-complete-transfer-connection-poo
and
https://forums.aws.amazon.com/message.jspa?messageID=532114#532353

Fix
http://stackoverflow.com/questions/29488106/java-lang-illegalstateexception-connection-pool-shut-down-while-using-spring-re


shorea:
I don't think this is the right fix. Each SDK client gets it's own connection pool that's managed internally. Setting this flag will prevent the connection pool from being cleaned up after the client is shutdown. It seems like the problem is your application (perhaps through the Flow framework) is trying to use an SDK client after it's been shutdown. If that's the case then the code should be modified to only shutdown the client after it's finished using it.

shorea:
Closing this PR. Feel free to re-engage if the problem persists.

Issue #952, opened 2016-12-20
calls to aws rest api may have side-effect of writing to System.err


davidmoten:
When attempting to read messages from a queue using aws-sdk any message returned is parsed for xml content. When we experience an internet outage we can get any old crap back from our corporate internet gateway saying that an outbound connection (the call to the aws rest api) timed out. Xml parsing is attempted and an error is thrown which we trap and then retry. All good so far.  

The trouble is that this particular scenario can bring about messages written to `System.err` before the error is thrown. This is because the sun xerces parser (when running on say Oracle JDK) that is used by the sdk has a fatal error in parsing and writes the error to `System.err`. Our application containers map `System.err` to log lines of level ERROR/SEVERE and our monitoring goes off. I don't know what other JDKs do.

Overriding xerces logging behaviour seems tricky so I thought I'd see if I can make this a problem to be solved by the team here seeing as calls to the aws rest api using the aws sdk can have annoying side effects (in terms of output on `System.err`).

Below is a little test that shows the log line appearing. It occurs because `loose.dtd` is malformed (!).

```java
import java.io.IOException;
import javax.xml.parsers.ParserConfigurationException;
import org.junit.Test;
import org.xml.sax.SAXException;
import com.amazonaws.util.XpathUtils;
import au.gov.amsa.util.net.AmsaConstants;

public class ErrorTest {

    @Test
    public void testXPathUtilsErrorLogging()
            throws SAXException, IOException, ParserConfigurationException {
        XpathUtils.documentFrom("<!DOCTYPE HTML PUBLIC \"-//W3C//DTD HTML 4.01 Transitional//EN\"\n"
                + "\"http://www.w3.org/TR/html4/loose.dtd\">\n" + "<html>\n");
    }

}
```
System.err looks like:
```
[Fatal Error] loose.dtd:31:3: The declaration for the entity "HTML.Version" must end with '>'.
```

muellerc:
Did you already had a look at https://xerces.apache.org/xerces2-j/faq-dom.html#faq-10?
There is a possibility to register its own ErrorHandler which could ignore this errors.
But to be able to use it, we have to extend XpathUtils first. Not sure whether this is a good idea for a corner case in my eyes...

davidmoten:
Excuse the delay, missed your response due to holidays. Yep register an ErrorHandler but don't expose it as an option in the API, just ensure that nothing gets written to stderr by the current `XpathUtils.documentFrom` method. Does this seem reasonable?

[...]

PR #985, opened 2017-01-13
Suppress stderr logging by XpathUtils.documentFrom

davidmoten:
as discussed in #952.

The default Xerces error handler outputs to stderr when it has a problem parsing xml. This can generate alerts for users of containers like glassfish because stderr is mapped to ERROR log lines by default.

This PR suppresses logging to stderr by overriding the Xerces error handler.

The PR includes a unit test that fails on the original code.

### DIFF
--- a/aws-java-sdk-core/src/main/java/com/amazonaws/util/XpathUtils.java
+++ b/aws-java-sdk-core/src/main/java/com/amazonaws/util/XpathUtils.java
@@ -21,6 +21,7 @@ import java.net.URL;
 import java.nio.ByteBuffer;
 import java.util.Date;
 
+import javax.xml.parsers.DocumentBuilder;
 import javax.xml.parsers.DocumentBuilderFactory;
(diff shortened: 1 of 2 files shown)


davidmoten:
I've updated this PR (and squashed commits) to 

* write debug level log messages on error and warning calls to the `ErrorHandler`
*  simplify the test
* use a singleton instance of the logging `ErrorHandler` via the Holder pattern (I'm not sure what your preferred idiom is for singletons).

zhangzhx:
Thanks for this contribution. We have approved the PR in the team with minor changes collapsing the holder class and making the ERROR_HANDLER as a field.

One more thing needs your confirm that are you submitting this change under Apache License?

davidmoten:
Updated PR making ERROR_HANDLER a field. Yes, I'm submitting this change under Apache License.

zhangzhx:
Thanks for the change. This will be released today!

davidmoten:
Great, love your quick release cycle, thanks.

PR #992, opened 2017-01-21
Using isNullOrEmpty method

sakky11:
The method "isNullOrEmpty" is already there so why not to use it instead of writing the same code again ?


varunnvs92:
Thanks for the PR. Can you confirm you are submitting this under Apache 2.0 license?

sakky11:
Yes, I confirm that I am submitting it under Apache 2.0 license.

varunnvs92:
Thanks, the PR has been merged.

PR #1015, opened 2017-02-07
Pass DynamoDb mapper config when using Table Mapper

vvondra:
This is to fix the following scenario

```java
     new DynamoDBMapper(
            dynamoDB,
            DynamoDBMapperConfig.builder()
                .withTableNameOverride(withTableNameReplacement(tableName))
                .build()
     ).newTableMapper(DomainObject.class);
```

In the scenario above, the TableMapper does not respect the mapper config, e.g. in this case the table name override. When loading the object, it uses the `@DynamoDbTable` annotation to find the table name instead of using the override table value.

Some methods already use `config`, but these PR makes the whole class consistent.


millems:
Sorry for the delay on this one.

Did we still want to integrate this change into 1.11.x? We're mostly heads-down on the 2.0 SDK, and we'll make sure this issue is fixed when we redesign the dynamo DB mapper as part of this issue: https://github.com/aws/aws-sdk-java-v2/issues/35

millems:
Closing due to no reply. Feel free to reopen if this is still something you want merged. Sorry again for the original delay.

vvondra:
@millems the PR is still valid and I have a workaround in place. there's no BC break and the PR still cleanly merges, anything you'd need to get added to have it considered?

PR #1016, opened 2017-02-08
support case-insensitive header field for HTTP response header

t2y:
According to RFC 2616 section 4.2 HTTP header fields should be processed case insensitive (http://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2):

> Each header field consists of a name followed by a colon (":") and the field value. Field names are case-insensitive.

I found that _ServerSideEncryptionHeaderHandler_ uses raw response header, then it causes an error with case-sensitive. Also this issue is not limited to only s3 service.

```java
public class ServerSideEncryptionHeaderHandler <T extends ServerSideEncryptionResult> implements HeaderHandler<T>{

    /* (non-Javadoc)
     * @see com.amazonaws.services.s3.internal.HeaderHandler#handle(java.lang.Object, com.amazonaws.http.HttpResponse)
     */
    @Override
    public void handle(T result, HttpResponse response) {
        result.setSSEAlgorithm(response.getHeaders().get(Headers.SERVER_SIDE_ENCRYPTION));
        result.setSSECustomerAlgorithm(response.getHeaders().get(Headers.SERVER_SIDE_ENCRYPTION_CUSTOMER_ALGORITHM));
        result.setSSECustomerKeyMd5(response.getHeaders().get(Headers.SERVER_SIDE_ENCRYPTION_CUSTOMER_KEY_MD5));
    }
}
```

* https://github.com/aws/aws-sdk-java/blob/master/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/internal/ServerSideEncryptionHeaderHandler.java#L31

Using _TreeMap_ solves it as the same as _ObjectMetadata.metadata_, I think.

* https://github.com/aws/aws-sdk-java/blob/master/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/model/ObjectMetadata.java#L58

## Reference

* https://github.com/aws/aws-sdk-java/pull/590

### DIFF
--- a/aws-java-sdk-core/src/main/java/com/amazonaws/http/HttpResponse.java
+++ b/aws-java-sdk-core/src/main/java/com/amazonaws/http/HttpResponse.java
@@ -24,6 +24,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.TreeMap;
 
 /**
  * Represents an HTTP response returned by an AWS service in response to a
@@ -37,7 +38,7 @@ public class HttpResponse {
     private String statusText;
     private int statusCode;
     private InputStream content;
-    private Map<String, String> headers = new HashMap<String, String>();
+    private Map<String, String> headers = new TreeMap<String, String>(String.CASE_INSENSITIVE_ORDER);
     private HttpContext context;
 
     /**


shorea:
We do proper case insensitivity in the getHeader that accepts the header name. https://github.com/t2y/aws-sdk-java/blob/afa72784c5966a1c89ba565a32e59961631c2d3c/aws-java-sdk-core/src/main/java/com/amazonaws/http/HttpResponse.java#L96-L103. This change looks good to me though as well, will help prevent future misuse of this interface.

t2y:
Thank you for reviewing. Using TreeMap might be more efficient than iterating large headers. At least, it reduce the maintenance cost of code base since we don't need as below.

```java
    public String getHeader(String headerName) {
        for (Map.Entry<String, String> header : headers.entrySet()) {
            if (header.getKey().equalsIgnoreCase(headerName)) {
                return header.getValue();
            }
        }
        return null;
    }
```

shorea:
Yep sounds good. +1

shorea:
Since we are using the tree map do you mind removing the logic in getHeader and just do a map lookup?

zhangzhx:
Thanks for the contribution. It will be included in the next release.

Issue #1038, opened 2017-03-01
AWS Lambda: LambdaInvokerFactory does not permit RequestStreamHandler implementation


rostskadat:
Hi,

Here is the scenario:
1- a Lambda function that implement the com.amazonaws.services.lambda.runtime.RequestStreamHandler interface
```java
public class MyRequestStreamHandler implements RequestStreamHandler {
    public void handleRequest(InputStream is, OutputStream os, Context context) throws IOException {
        int letter;
        while((letter = inputStream.read()) != -1)
        {
            outputStream.write(Character.toUpperCase(letter));
        }
    }
}
```
2- a java class that tries to call that Lambda function:

```java
public class MyLambdaConsumer {

    public interface MyLambdaFunction {
       @LambdaFunction(functionName = "MyRequestStreamHandler")
       void handleRequest(InputStream is, OutputStream os);
    } 

    public void consume(AWSLambdaClient lambdaClient, InputStream is, OutputStream os) {
        LambdaInvokerFactory.Builder builder = LambdaInvokerFactory.builder();
        MyLambdaFunction lambdaFunction = builder.lambdaClient(lambdaClient).build(MyLambdaFunction.class);
        lambdaFunction.handleRequest(is, os);
   }
}
```

=> This will fail with a 

```
Caused by: com.amazonaws.services.lambda.invoke.LambdaSerializationException: LambdaFunctions take either 0 or 1 arguments
	at com.amazonaws.services.lambda.invoke.LambdaInvokerFactory$LambdaInvocationHandler.validateInterfaceMethod(LambdaInvokerFactory.java:235)
```

=> I would expect that the Builder would allow to create a LambdaFunction stub from a function that implements the RequestStreamHandler. I believe that the check in LambdaInvocationHandler.validateInterfaceMethod(LambdaInvokerFactory.java:235) is incorrect:

```java
    if (args != null && args.length > 1) {
        throw new LambdaSerializationException("LambdaFunctions take either 0 or 1 arguments");
    }
```

shorea:
With streaming though it wouldn't make sense to have an OutputStream as the second parameter. You'd want the input stream of the data coming back from the function.

rostskadat:
Hum, I understand. However, how does AWS knows that the second parameter of the LambdaFunction implementing the ``RequestStreamHandler`` should be streamed to the InputStream return by the local interface's method?

shorea:
[...]

PR #1042, opened 2017-03-01
Fixing issue #1038 by removing excessively restricting check.

rostskadat:


varunnvs92:
As per shorea@ comments in #1038, it seems the check is valid and should not be removed. I will sync up with shorea@ and see what is the right solution for this. Closing the PR.

PR #1045, opened 2017-03-02
Don't download keys from directory that match a given regex pattern

daniel-puschila:
TransferManager's downloadDirectory() should allow a selective download of the keys.


varunnvs92:
What is the use case of this feature in your application? Is it possible for you to upload the files (that match regex) to different bucket?

We generally don't add new features to high level utilities unless there is a strong use case and is requested by multiple customers. If we add this feature to download directory, there might be requests to add the same feature for copy and upload operations which we don't want to do unless there is a strong use case. Thank you.

kiiadi:
If we were to take this feature in it probably makes more sense to use some sort of predicate instead of a straight-up regex. Similar to the built-in [FilenameFilter](https://docs.oracle.com/javase/7/docs/api/java/io/FilenameFilter.html) - that will give customers more flexibility for how they filter.

varunnvs92:
@daniel-puschila 
As mentioned earlier, we are not ready to accept PRs for new features unless many customers ask for it. We are planning to release a new version of the SDK with better features later this year. I can add it to the backlog. If you like, you can create a new issue in the repo and tag it with Feature Request. Ask people to +1 it if they are interested in the feature which will help us prioritize.

Issue #547, opened 2015-10-29
DynamoDB: Support getter and setter less fields.


gkristjansson:
It would be super nice not having to define getters and setters for POJO when using mapping. The annotations are already allow to be placed on fields, but requires getters and to present. Often the encapsulation argument does not always clearly apply, since in lots of cases the type that you are using for storage is just used for that and is not accessible else where in the code base.

zhangzhx:
Could you give me a concrete example that why you want a getter and setter less fields? The POJO is not only used to store data retrieved from DynamoDB, it also could be used to modify data and save to DynamoDB. The getter and setter are essential to make that happen.

Dimes:
Getters and setters are compile time contracts about what is allowed on an object. The DynamoDBMapper is using reflection to set the internal state of a class. The mapper has no need to respect the compile-time contract, and requiring the use of setters makes perfectly valid application logic impossible. For instance, I should be able to model a persisted class with `final` instance variables, but `final` fields are impossible to use with setters. In some sense, DynamoDB is used for persisting application state. Why should any persistence library force the application to conform to a certain type of API when Java's reflection API already contains the required functionality for persisting state?

Can you give an example when getters and setters need to be used?

kdabir:
@Dimes I totally agree with you on this. I usually try to mark as many fields `final` as possible and get required data in constructor(s) or through some kind of builders, but try to keep the model as immutable as possible. 

How I am dealing with this is creating dumb model classes for dynamo mapper (under a different pacakge), and then mapping them to/from my richer Domain model.

shorea:
You can also define a read only interface and return that to your domain logic. The setters would only be visible and used by your mapping layer.

kepkap:
Hi guys,
+1 vote for this issue. 

having explicit public setters on persisted and presumably rich-modelled entity is something you want to avoid. This constraint opens the door to anemic domain model antipattern.  

I am totally fine with java bean nature for simple entities though. 

Still evaluating dynamodb for our project (migrating from mongodb), so I am up for this  issue in advance. 

thnx

rage-shadowman:
IMO, immutability is a necessity in any code that runs more than 1 thread. This way we can create immutable model objects that we can pass around without worrying about any bad actors modifying the data from underneath us, since they will have to build their own snapshot and leave ours alone.

[...]

PR #1051, opened 2017-03-06
Issue #547: Field level POJO accessors

mgreenwood1001:
Immutable entity support (https://github.com/aws/aws-sdk-java/issues/547) - I could see using a pattern where-in the domain entity is immutable for "store" behavior.  Although, it seems unlikely as the predominate case as I'd likely have some form of transfer object to marshal state to/from the DB where-in I have setters/getters.

Just an option for an implementation that solves this concern as I myself might want to do something similar in some of our use-cases.

Thought this was an interesting problem to solve, here's one possible implementation.     Might consider re-using reflectutils rather than directly using java.lang.*... Also might not want to invoke field.set when setter not available...

Tests are needed by this implementation, willing to add them if this is worthy of eventually merging.


millems:
Sorry for the delay on this.

We've got immutable model support as the first suggestion for V2 of the mapper: https://github.com/aws/aws-sdk-java-v2/issues/35

We're mostly code frozen on major features being added to 1.11.x's mapper. We'll make sure to think about this problem as we redesign V2. Sorry we weren't able to integrate your contribution. We appreciate it.

PR #1054, opened 2017-03-07
Correctly unwrap ExecutionExceptions

### LINKED ISSUE #1046: Wrong Exception in TransferManager#download
When using the TransferManager's download method to download an object in a parallelised way and something fails the client will be presented with an AmazonClientException instead of the expected AmazonS3Exception.

Steps to repro (sorry I can't provide a JUnit test for that due to the parallelism):

* upload an object which can be downloaded with a parallel request
* set a breakpoint DownloadCallable#downloadInParallel (line 188)
* initiate the request to download the object
* if you hit the breakpoint, delete the object manually from S3 (this emulates another client which deletes the object)
* resume the breakpoint
* DownloadCallable#combineFiles will throw an ExecutionException with the wrapped AmazonS3Exception.
* this is caught in DownloadCallable#call and afterwards is rethrown.
* Since DownloadCallable is also executed in a ExecutorService it will be wrapped again in an ExecutionException. We now have (ExecutionException -> ExecutionException -> AmazonS3Exception)
* The Exception is then caught in AbstractTransfer#unwrapExecutionException which will unwrap only the top most exception and the does a check if it is an AmazonClientException (which it is not and thus wrap everything in AmazonClientException which leads to AmazonClientException -> ExecutionException -> AmazonS3Exception)
* Next exception handling has to be done by the client (i.e. transferManager.download(xyz).waitForCompletion(). The exception presented to the client is an AmazonClientException where the expected exception should be only the AmazonS3Exception

We do expect the Exception to be the AmazonS3Exception b/c we have some special handling for that case. In the single download case this works as expected but fails in the parallel download. I guess the right fix would either be to implement AbstractTransfer#unwrapExecutionException to unwrap until all ExecutionExceptions are gone or in the exception handling of DownloadCallable to unrwrap the ExecutionException already there.

smigfu:
CLOSES #1046

### DIFF
--- a/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/transfer/internal/AbstractTransfer.java
+++ b/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/transfer/internal/AbstractTransfer.java
@@ -276,8 +276,12 @@ public abstract class AbstractTransfer implements Transfer {
      * @return The root exception that caused the specified ExecutionException.
      */
     protected AmazonClientException unwrapExecutionException(ExecutionException e) {
-        Throwable t = e.getCause();
-        if (t instanceof AmazonClientException) return (AmazonClientException)t;
+        Throwable t = e;
+        while (t.getCause() != null && t instanceof ExecutionException) {
+            t = t.getCause();
+        }
+
+        if (t instanceof AmazonClientException) return (AmazonClientException) t;
         return new AmazonClientException("Unable to complete transfer: " + t.getMessage(), t);
     }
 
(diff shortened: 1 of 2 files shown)


kiiadi:
Great! Thanks for adding the test too.

Are you OK to submit this under Apache 2.0 license? Assume so since you've added the header in the test file just wanted to check.

smigfu:
Yes, I am absolutely fine with that.

smigfu:
@kiiadi : Is there anything else you require from me or can this PR be merged?

varunnvs92:
Sorry for the delay. We will include this in our next release (today or tomorrow). Thank you for your contribution.

PR #1070, opened 2017-03-14
Doc fix for `ClientConfiguration::getProxyUsername`

rhenning:
* this method returns the `http(s).proxyUser` system property,
  not `http(s).proxyUsername`.

Hello. I believe this small documentation update is warranted based on the content of the `getProxyUsernameProperty()` method.  The system property is correct in code but incorrect when reading the documentation.


dagnir:
Thanks! LGTM. This will be included in our next release.

PR #1076, opened 2017-03-21
AmazonS3Client#copyObject should fire transfer events

jakeab:
The COPY action is not firing events similar to GET or PUT. The TransferManager fires events for multipart copies, and hints that AmazonS3Client should be responsible for non-multipart copies.

https://github.com/aws/aws-sdk-java/blame/master/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/transfer/internal/CopyMonitor.java#l=152

### DIFF
--- a/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/AmazonS3Client.java
+++ b/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/AmazonS3Client.java
@@ -1837,6 +1837,9 @@ public class AmazonS3Client extends AmazonWebServiceClient implements AmazonS3 {
          * remote server thinks there's more data to pull.
          */
         setZeroContentLength(request);
+
+        final ProgressListener listener = copyObjectRequest.getGeneralProgressListener();
+        publishProgress(listener, ProgressEventType.TRANSFER_STARTED_EVENT);
         CopyObjectResultHandler copyObjectResultHandler = null;
         try {
             @SuppressWarnings("unchecked")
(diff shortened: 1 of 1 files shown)


jakeab:
Hi,

Do you need more information from me, or have other feedback? I'm having to work around this inconsistency in my application.

Thanks,
Jake

millems:
Sorry about the delay on this. The documentation isn't very clear on defining a transfer.

Transfers are defined as streaming files to/from an AWS service. Because this isn't streaming a file (or InputStream), the lack of transfer events is expected. Feel free to check out the other event types to see if they meet your need better, like `CLIENT_REQUEST_STARTED_EVENT`.

jakeab:
I believe what you describe is not what happens in transfer manager. In any case, it is inconsistent between a synchronous copy using the AmazonS3Client and the asynchronous with TransferManager. The other events don't really meet my needs. Ideally, this would be consistent across downloads, uploads and copies. Thanks.

millems:
Oh yeah, that's really weird. We talked it out and agree that it's even really confusing. Transfer events are only really published when we can provide incremental updates about the progress of processing the request. 

So uploads and downloads are fair game because we know the number of bytes as they transfer. Copy parts are fair game because we know the size of each part and can tell when each one finishes. Copy with no parts aren't published because we don't know anything other than we started the copy and finished the copy.

We agree that's a pretty dumb distinction, but we're hesitant to change it in 1.11.x because we're not sure how customers are using these events. If they're using the same distinction, adding them to a single copy will break them because there's no progress.

We're going to come up with a more consistent and useful set of events for V2. Would you mind describing your use-case on our V2 issue so we can make sure it's great when we implement progress listeners there? https://github.com/aws/aws-sdk-java-v2/issues/25

PR #1091, opened 2017-03-29
The file MANIFEST.MF does not have required information

frekele:
The file MANIFEST.MF does not have required information.

For example, I'd like to know what version of the jar is running in my application at runtime.

```
AmazonS3Client.class.getPackage().getImplementationVersion();
```
But currently I'm getting the info with value: null, Because, does not contain anything into MANIFEST.MF.

Please, include this in the project. 

Thank you.


frekele:
Based on the file:
https://github.com/jboss/jboss-parent-pom/blob/master/pom.xml

kiiadi:
@frekele our pom files are auto-generated - I can look at add the goals you're suggesting to get more information in the MANIFEST.MF but in the meantime - can you get what you need from `AmazonS3Client.getProtectionDomain().getCodeSource().getLocation();`
?

frekele:
```
AmazonS3Client.class.getProtectionDomain().getCodeSource().getLocation();
```
This line of code, returns the file path with the name, it will be necessary to treat the String to get the value. But I do not think I should trust the filename.

It would be great to have the default information into MANIFEST.MF.
```
Implementation-Version:  
Implementation-Title: 
Implementation-Vendor-Id:  
Specification-Title: 
Specification-Version: 
...... and others.
```
Thanks for the feedback.

Let's wait this.

kiiadi:
I've opened this as a feature request - please feel free to up-vote it!

PR #1124, opened 2017-04-21
Fix NPE fault in Regions.fromName

amcp:
This change reverses the lhs and rhs of the .equals method used in Regions.fromName to avoid an NPE. Also adds relevant tests.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,6 @@
 target/
 *settings.xml
+.idea
+*iml
+*~
+.*~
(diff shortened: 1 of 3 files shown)


varunnvs92:
Looks good to me. Can you confirm that you are submitting under Apache 2.0 license?

amcp:
@varunnvs92 confirmed apache 2.0

varunnvs92:
Thanks for your contribution. This will be included in today's release.

PR #1125, opened 2017-04-24
Clear last used provider when refreshing AWSCredentialsProviderChain

mfenniak:
Currently `AWSCredentialsProviderChain`'s cache of the last used credential provider is maintained across `refresh()` invocations in the provider chain.  This means that after you've accessed the credentials once, `refresh()` effectively only refreshes the credential provider that was previously accessed; you can never switch to one earlier in the chain via a refresh.

This PR clears the cached last provider in a refresh.

### DIFF
--- a/aws-java-sdk-core/src/main/java/com/amazonaws/auth/AWSCredentialsProviderChain.java
+++ b/aws-java-sdk-core/src/main/java/com/amazonaws/auth/AWSCredentialsProviderChain.java
@@ -131,9 +131,35 @@ public class AWSCredentialsProviderChain implements AWSCredentialsProvider {
         throw new SdkClientException("Unable to load AWS credentials from any provider in the chain");
     }
 
+    /**
+     * Forces this credentials provider to refresh the credentials of all
+     * credential providers in its chain.
+     *
+     * If configured to reuse the last successful credential provider, the
+     * cache of the last successful credential provider will be retained.
+     */
     public void refresh() {
+        refresh(false);
+    }
+
+    /**
+     * Forces this credentials provider to refresh the credentials of all
+     * credential providers in its chain.
+     *
+     * If configured to reuse the last successful credential provider, the
+     * cache of the last successful credential provider may be cleared by
+     * providing the resetLastUsedProviderCache flag.
+     *
+     * @param resetLastUsedProviderCache
+     *            Whether or not to clear the last successful credential
+     *            provider; a true value will clear it.
+     */
+    public void refresh(boolean resetLastUsedProviderCache) {
         for (AWSCredentialsProvider provider : credentialsProviders) {
             provider.refresh();
+        }
+        if (resetLastUsedProviderCache) {
+            lastUsedProvider = null;
         }
     }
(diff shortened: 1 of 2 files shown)


kiiadi:
I don't think we'll be able to accept this as is because it's a change in behaviour - existing customers may be relying on the fact that only the 'last used' provider gets refreshed. Couple of possibilities that I see we could do:
1. Introduce a new method on the chain that does this (e.g `clearLastUsed` or even an overload of `refresh()` that takes a boolean to determine whether last used is cleared).
2. Introduce a new implementation of the `AWSCredentialsProviderChain` that does this.
3. Introduce a new constructor on `AWSCredentialsProviderChain` that takes a `clearLastUsedOnRefresh` boolean flag that controls this behaviour - the default would be `false`.

shorea:
What's your use case?

mfenniak:
Thanks for the feedback, @kiiadi.  I've updated this pull request, following the approach of making the cache clear behavior optional with a boolean flag.

@shorea: My use-case is a tool that may be configured to use instance profile credentials to download preconfigured credentials from S3, and write them to ~/.aws/credentials.  I'd like to refresh the credential provider chain between these two steps, and in that configuration, the active credential provider would switch from one provider to another.

shorea:
That sounds like a good candidate for implementing a custom AWSCredentialsProvider instead of trying to use a chain to achieve this behavior. Chains are more geared towards supporting running the same application to multiple environments (i.e. EC2, Lambda, ECS, local dev) where the credentials may be sourced from a different location depending on the env but they never change locations once sourced. 

Also do you need to persist the credentials to the ~/.aws/credentials file or can you just cache them in memory? Will other processes use those credentials?

mfenniak:
@shorea: That would be an alternative implementation, sure.  But it would require a lot more development work than just downloading a file and calling refresh on the default credential provider.  The tool under development here is running in a docker container, so downloading to ~/.aws/credentials is effectively not persistent.

I get that the use-case is unusual.  I'm not really asking this SDK to support that use case; I have a workaround in place already, albeit it is ugly.

But it seems justifiable to me that there be a mechanism to clear a cache like the one in `AWSCredentialsProviderChain`.  It seems to fit well with the `refresh` method, which is documented as 'Forces this credentials provider to refresh its credentials.'  I think this PR adds the functionality in a clear and compatible manner.  If it's not a change you'd like to adopt, I understand. 😀

Here's a workaround, for reference:

```java
AWSCredentialsProviderChain defaultAWSCredentialsProviderChain =
    DefaultAWSCredentialsProviderChain.getInstance();
defaultAWSCredentialsProviderChain.setReuseLastProvider(false);
defaultAWSCredentialsProviderChain.refresh();
try { defaultAWSCredentialsProviderChain.getCredentials(); } catch (Exception e) { }
defaultAWSCredentialsProviderChain.setReuseLastProvider(true);
```

millems:
Sorry to say that this use-case is a bit too specific for us to include in the general SDK. We may revisit this decision if more customers need the feature.

Issue #1029, opened 2017-02-22
None of the SDK objects override equals or hashcode


HunterSherms:
This makes it nearly impossible to build useful unit tests for methods that call SDK methods since you can't verify the objects passed to them.

For a random example (though this applies everywhere) say my code wants to list objects in an S3 bucket:

```
        return s3Client.listObjectsV2(new ListObjectsV2Request()
                .withBucketName(MY_BUCKET_NAME)
                .withDelimiter("/")
                .withPrefix(MY_PREFIX);
```

It is effectively impossible for me to verify in a unit test that listObjectsV2 was called with a ListObjectsV2Request object that contains MY_BUCKET_NAME and MY_PREFIX without extending every SDK object you use.

This could be solved by adding an equals and hashcode method to each object in this SDK.

zhangzhx:
Thanks for reaching out to us. Not like other clients, the S3 client is handwritten and we don't have equal, hashCode methods for the Request objects.

This is a nice feature request and feel free to send us a pull request and we are glad to take a look.

HunterSherms:
Ok, will do.

debora-ito:
Fixed with an internal change as mentioned in https://github.com/aws/aws-sdk-java/pull/1133#issuecomment-395896362.

PR #1133, opened 2017-04-28
Adding Equals and Hashcode to GetObjectRequest and S3ObjectBuilder

HunterSherms:
Adding Equals and Hashcode to GetObjectRequest and S3ObjectBuilder to improve the ability to unit test methods that create and use these objects.

https://github.com/aws/aws-sdk-java/issues/1029

### DIFF
--- a/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/model/GetObjectRequest.java
+++ b/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/model/GetObjectRequest.java
@@ -16,6 +16,7 @@ package com.amazonaws.services.s3.model;
 
 import java.io.Serializable;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Date;
 import java.util.List;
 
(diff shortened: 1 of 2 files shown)


aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/model/GetObjectRequest.java
```diff
@@ -1008,4 +1008,47 @@ public GetObjectRequest withS3ObjectId(S3ObjectId s3ObjectId) {
...
+        if (this == object) return true;
+        if (object == null || getClass() != object.getClass()) return false;
+        if (!super.equals(object)) return false;
+
+        GetObjectRequest that = (GetObjectRequest) object;
+
+        if (isRequesterPays() != that.isRequesterPays()) return false;
+        if (!s3ObjectIdBuilder.equals(that.s3ObjectIdBuilder)) return false;
+        if (!java.util.Arrays.equals(getRange(), that.getRange())) return false;
+        if (!getMatchingETagConstraints().equals(that.getMatchingETagConstraints())) return false;
```
kiiadi: Might need a null check for `getMatchingETagConstraints`

aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/model/GetObjectRequest.java
```diff
@@ -1008,4 +1008,47 @@ public GetObjectRequest withS3ObjectId(S3ObjectId s3ObjectId) {
...
+        if (!nonmatchingEtagConstraints.equals(that.nonmatchingEtagConstraints)) return false;
```
kiiadi: Is there a reason you sometimes reference the property directly `!nonmatchingEtagConstraints.equals(that.nonmatchingEtagConstraints)` and you sometimes use the getter `getModifiedSinceConstraint() != null ? !getModifiedSinceConstraint().equals(that.getModifiedSinceConstraint()` ?

aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/model/GetObjectRequest.java
```diff
@@ -1008,4 +1008,47 @@ public GetObjectRequest withS3ObjectId(S3ObjectId s3ObjectId) {
         setS3ObjectId(s3ObjectId);
         return this;
     }
+
+    @Override
+    public boolean equals(Object object) {
+        if (this == object) return true;
+        if (object == null || getClass() != object.getClass()) return false;
```
kiiadi: Any reason you don't use `instanceof` instead of `getClass()`? It seems like it would cover the null scenario; it's also more consistent with the other equals operators in the SDK.

HunterSherms: instanceof returns true for subtypes

I would personally be surprised if an object of type A compared to an object of type B extends A returned true. Is this the desired behavior?

aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/model/GetObjectRequest.java
HunterSherms: Good catch... I used a template for these and the getter has the capitalization "ETag" while the property is "Etag". Fixing.

aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/model/GetObjectRequest.java
kiiadi: I think wars have been fought over this topic already so I don't really want to go into the pros/cons of each approach. 😃 Josh Bloch (Effective Java) recommends using `instanceof` ([see here](http://www.artima.com/intv/bloch17.html)) and it's what the rest of the SDK does.

If we make the `equals`/`hashCode` methods final then we somewhat mitigate the issue because we can preserve the `a.equals(b) == b.equals(a)` behaviour even across subclasses.

`instanceof` is consistent with the rest of the SDK so I'd prefer to stick to that for now as it's less surprising.

HunterSherms: ok will update

kiiadi: Thanks - sorry to be a stickler but when faced with these sorts of decisions I usually opt for consistency - it's important that we do our best to ensure the SDK behaves in a predictable way.

HunterSherms: updated

millems:
This can't be merged because of failed checks, but we've implemented this locally and it should go out with our next release, some time next week.

PR #1137, opened 2017-04-28
Add support for session tokens in SystemPropertiesCredentialsProvider

greglook:
This is a small change to add support to the `SystemPropertiesCredentialsProvider` for reading STS session tokens from the JVM properties. Currently other providers such as `EnvironmentVariableCredentialsProvider` support STS tokens, but there's no way to change the environment at runtime.

I want to be able to run a boot-up process to assume an IAM role with STS and 'install' the keys into the system properties so that any library using the AWS SDK works seamlessly and I don't need to thread a custom credentials provider implementation throughout the system.


shorea:
LGTM. Can you confirm you are submitting under the Apache 2.0 license?

greglook:
> Can you confirm you are submitting under the Apache 2.0 license?

Yes, I release any and all copyright claims on the contributed work. I can sign a CLA if needed.

shorea:
Thanks for confirming, merging this in. Will be released sometime next week, probably monday.

PR #1155, opened 2017-05-10
Correct javadoc for setExpirationTimeRuleId: for internal use only

apeeters:
Add javadoc similar to setExpirationTime.


millems:
This can't be merged because of failed checks, but we've implemented this locally and it should go out with our next release, some time next week.

PR #1180, opened 2017-06-02
updating SES sample code to use builder instead of deprecated constructors

soo-aws:
Got a ticket from internal customer regarding this sample


NoriakiHoriuchi:
@varunnvs92 Hi! What's the status now?

millems:
This can't be merged because of conflicts, but we've implemented this locally and it should go out with our next release, some time next week.

PR #1198, opened 2017-06-16
fixing doc typos and spelling

soo-aws:
minor fixes from customer tickets.

The sentence I took out was confusing customers because the "Examples" didn't link to anything.


dagnir:
Thanks @soo-aws. The change to the mapper looks good but the javadoc from `AmazonRoute53` is taken from the service models so they would need to fixed there.

varunnvs92:
Made changes to mapper internally which will be included in todays releaes. As @dagnir pointed out, Route 53 changes should be fixed on service end.

Issue #1215, opened 2017-06-30
Performance Bottleneck: TransferManager downloadDirectory 


bawejakunal:
In the `TransferManager.downloadDirectory` a [depth-first-search has been implemented](https://github.com/aws/aws-sdk-java/blob/d65733f87550d74a042e30f630c4b9033b6e3372/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/transfer/TransferManager.java#L1210) where the list objects request is formed as below:

```
ListObjectsRequest listObjectsRequest = new ListObjectsRequest().withBucketName(bucketName)
                            .withDelimiter(DEFAULT_DELIMITER).withPrefix(prefix);
```
Well when we want to download all of the files of a directory based on a `KeyFilter` then why do we need a `ListObjectsRequest.withDelimiter`. Instead we could just have had:

```
ListObjectsRequest listObjectsRequest = new ListObjectsRequest().withBucketName(bucketName)
                            .withPrefix(prefix);
```

I have more than 13000 files which I can list easily in 14 API calls with the latter method. But the `downloadDirectory()` call makes 27000 API calls for the same because at each API call it goes down just 1 level of virtual directory.

Is there a particular justification for such implementation using `withDelimiter(DEFAULT_DELIMITER)` because as far as I see all possible filtering can be achieved through `KeyFilter` and `keyPrefix`.

longi5rui:
Hey guys, any update on this issue? I always got `java.lang.OutOfMemoryError` when using TransferManager `downloadDirectory` API to download a directory with 3 million files.

This is a issue also for TransferManager `uploadDirectory` API but I use the approach @millems suggested which is I use `DirecotryStream` + Multi-threading my own to upload directory with large number files.

millems:
Sorry that this was ignored for so long. We're nervous about making non-trivial changes for 1.11.x at this time, because we've found that even seemingly-safe changes can affect a handful of users and pull us away from development on 2.x of the SDK.

We'll incorporate this feedback while building the SDK 2.x.

zoewangg:
Hi all, just a quick update, this feature has been implemented in the transfer manager in Java SDK v2. Note that it's in Developer Preview. You can check out sample code here https://github.com/aws/aws-sdk-java-v2/tree/master/services-custom/s3-transfer-manager. 

As always, we welcome feedback, bug reports and feature requests on the https://github.com/aws/aws-sdk-java-v2 repo!:)

PR #1223, opened 2017-07-07
performance bottleneck fix for issue #1215

bawejakunal:
A depth first directory traversal leads to unnecessary network calls to S3 API and drastically reduces performance. In case of recursive download with prefix, simple listing and filtering of object keys should be sufficient.

Details:  https://github.com/aws/aws-sdk-java/issues/1215


millems:
Sorry that this PR was ignored for so long. We're nervous about taking non-trivial pull requests for 1.11.x at this time, because we've found that even seemingly-safe changes can affect a handful of users and pull us away from development on 2.x of the SDK.

We'll try our best to incorporate this feedback when we develop the 2.x transfer manager.

bawejakunal:
Haha thanks :D for finally getting around to this good know this will be up in next SDK version

RIP for all 1000s of API calls from existing SDK clients

PR #1255, opened 2017-07-26
add support for S3ContextAwareCredentialsProvider (Requesting Feedback/Comments)

zarfide:
Hello there,

We are *constantly* dealing with multi-account permissions related issues with S3 in our development environment.  We have hundreds of developers working over dozens of accounts who are very good at using the wrong credential to try to read/write to a particular bucket with the wrong credential or the wrong object ownership settings.  It's a seemingly never-ending pain point for us.

We would *very* much like a solution for which we can provide a "smarter" credential source which automatically sources the most relevant credential for interacting with a particular bucket/key and thus by default also provides the most correct ownership of a particular object.  The following PR is I think the minimum viable change which allows this functionality, but I'm open to some other approach.   Basically it would allow us to provide our own logic for sourcing the credentials; we just need the hooks to send back the bucket/key name so we can make the appropriate decisions in our own logic.

I considered just having a wrapper over all of the AmazonS3Client methods and adapting them into the equivalent method that takes a AmazonWebServiceRequest (in which we could also automatically set the provider), but this seems like a lot of boilerplate code (that would also be subject to breakage every time a new method is created) for a change that can logically be made in just a few places.

I reiterate that this is a huge pain point for us and something I'd really like to get addressed without having to fork the SDK and then constantly have to backport updates.

I also noticed that there do not seem to be any S3-related tests or am I just not looking in the right place? I'd be happy to add the relevant tests as well but I'm not really sure what the paradigm is for testing the s3 package.

Thanks for reading,
Jay Zarfoss (Platform Security @ Netflix)

### DIFF
--- a/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/AmazonS3Client.java
+++ b/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/AmazonS3Client.java
@@ -2885,7 +2885,11 @@ public class AmazonS3Client extends AmazonWebServiceClient implements AmazonS3 {
             // delegate directly to it.
             ((Presigner) signer).presignRequest(
                     request,
-                    CredentialUtils.getCredentialsProvider(request.getOriginalRequest(), awsCredentialsProvider).getCredentials(),
+                            CredentialUtils.getCredentialsProvider(request.getOriginalRequest(),
+                                    awsCredentialsProvider instanceof S3ContextAwareCredentialsProvider ?
+                                            ((S3ContextAwareCredentialsProvider) awsCredentialsProvider).getCredentialsProvider(bucketName, key) :
+                                            awsCredentialsProvider
+                            ).getCredentials(),
                     req.getExpiration()
             );
         } else {
@@ -3603,7 +3607,11 @@ public class AmazonS3Client extends AmazonWebServiceClient implements AmazonS3 {
         resourcePath = resourcePath.replaceAll("(?<=/)/", "%2F");
 
         new S3QueryStringSigner(methodName.toString(), resourcePath, expiration)
-            .sign(request, CredentialUtils.getCredentialsProvider(request.getOriginalRequest(), awsCredentialsProvider).getCredentials());
+            .sign(request, CredentialUtils.getCredentialsProvider(request.getOriginalRequest(),
+                    awsCredentialsProvider instanceof S3ContextAwareCredentialsProvider ?
+                            ((S3ContextAwareCredentialsProvider) awsCredentialsProvider).getCredentialsProvider(bucketName, key) :
+                            awsCredentialsProvider
+            ).getCredentials());
 
         // The Amazon S3 DevPay token header is a special exception and can be safely moved
         // from the request's headers into the query string to ensure that it travels along
@@ -4221,7 +4229,10 @@ public class AmazonS3Client extends AmazonWebServiceClient implements AmazonS3 {
                         new S3V4AuthErrorRetryStrategy(buildDefaultEndpointResolver(getProtocol(request), bucket, key)));
             }
 
-            executionContext.setCredentialsProvider(CredentialUtils.getCredentialsProvider(request.getOriginalRequest(), awsCredentialsProvider));
+            executionContext.setCredentialsProvider(CredentialUtils.getCredentialsProvider(request.getOriginalRequest(),
+                    awsCredentialsProvider instanceof S3ContextAwareCredentialsProvider ?
+                            ((S3ContextAwareCredentialsProvider) awsCredentialsProvider).getCredentialsProvider(bucket, key) :
+                            awsCredentialsProvider));
 
             validateRequestBeforeTransmit(request);
             response = client.execute(request, responseHandler, errorResponseHandler, executionContext);
--- a/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/S3ContextAwareCredentialsProvider.java
+++ b/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/S3ContextAwareCredentialsProvider.java
@@ -1,0 +1,15 @@
+package com.amazonaws.services.s3;
+
+import com.amazonaws.auth.AWSCredentialsProvider;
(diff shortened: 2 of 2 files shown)


millems:
Would it be sufficient for you to implement a `RequestHandler2` with a `beforeExecution` method that looks at the request and picks which credential provider to use with `setRequestCredentialsProvider`?

Quick and dirty example:
```Java
public class MyCredentialProviderRequestHandler extends RequestHandler2 {
    @Override
    public AmazonWebServiceRequest beforeExecution(AmazonWebServiceRequest request) {
        if (request instanceof GetObjectRequest) {
            GetObjectRequest getObjectRequest = (GetObjectRequest) request;
            String bucketName = getObjectRequest.getBucketName();
            AmazonWebServiceRequest clone = request.clone();
            clone.setRequestCredentialsProvider(credentialsProviderFor(bucketName));
            return clone;
        }
        
        return request;
    }
}
```

zarfide:
It's not a bad idea; I think it suffers from the same general issue as wrapping all the uses of `AmazonS3Client` though which is that I'd need an if block checking for every possible `AmazonWebServiceRequest` which may have a bucket/key possibly in it, and I'd need to keep that code in sync with whatever additional `AmazonWebServiceRequest` types got added over time.

millems:
That's a fair point. It feels a bit out there to ask you to reflectively check for a method that happens to be named `getBucketName` or `getKey`. 

Before I make a different suggestion, what's your desired behavior for operations that involve a source and destination bucket/key? Eg. copy requests. Would you want to make decisions based on either piece of information (source or destination)?

zarfide:
Good question; I honestly hadn't thought of copy requests.  Are there any other examples with multiple buckets for one request?

Probably in that case we care the most about the target/destination bucket, since the biggest pain point has to do with people writing into a bucket where target bucket is owned by account A and they write into an object owned by account B.  Our plan was to assumeRole into account A for them and let the write proceed without any other special object-level permissions.

millems:
Source/Destination requests are only `CopyObject` and `CopyPart`. Let me bounce some ideas around on my end and get back to you.

zarfide:
Just thinking out loud, perhaps a cleaner refactoring would be to make a new subclass of `AmazonWebServiceRequest`... something like `S3ContextAwareWebServiceRequest` that is guaranteed to have a source/target bucket/key name and move all of the relevant S3 web services requests to be children of the new `S3ContextAwareWebServiceRequest` class.  It would be a much bigger code change up front, but the request-handler approach becomes much cleaner since it would only have to check for one object type.    At that point you'd just have to make sure that any new request were extended from `S3ContextAwareWebServiceRequest` as appropriate.

It would still be slightly less desirable on this end since we'd have to make sure everyone was adding the relevant request handler (it was hard enough getting everyone to setup credentials properly) but it would be a marked improvement from the ad hoc request construction we're doing now.

millems:
S3 is a really special use-case right now in 1.11 in that we write the client by hand and have the ability to use inheritance/interfaces. Unfortunately that doesn't extend to other services or 2.0 of the SDK. We're hesitant to suggest something that only works in 1.11's S3.

After talking to the rest of the team we could only really come up with 2 solutions that fits your requirements and scales to other services and future SDK versions:

**Option 1:** Create a client provider that your customers use to retrieve a client based on the resource being manipulated that is pre-configured with the correct credentials. This sounds non-ideal because your customers would have to make code changes.

```Java
S3 client = S3ClientRepository.clientFor(bucket, key);
client.getObject(new GetObjectRequest().withBucketName(bucket).withKey(key))
```

**Option 2:** Create a message handler that assumes the S3 service team will always use the "bucketName" and "key" naming on their fields and use reflection to invoke that method or load that bean property. 

As far as getting your customers to actually use it, that's a lot easier with message handlers. The SDK scans the classpath for files named `com/amazonaws/global/handlers/request.handler2s`. All message handlers listed in this file (one per line) are added to all service clients your customer creates. If you ship your message handler implementation with this file, your customers only need to add a dependency on your library to get it loaded in automatically.

zarfide:
I think the bigger issue with Option 1 would be getting developers used to the paradigm that an S3Client is single use.  

As for Option 2, honestly it feels very much like a hack to me although I think it will generally meet our needs if we have some level of guarantee re: the `bucketName` and `key` fields.  If that's the best option I will propose that among the team here.  I'm not sure if we would go the classpath scanning route or ask devs to use a wrapper that installed the handler directly... classpath scanning has its own demons when dealing with large projects.

I feel like I should add that while this problem is exacerbated in S3 because of the differences in bucket/object ownership, other services have the same issue.  For example, we're currently in the process of putting this logic into our Route53 wrapper to do an assume role depending on the hosted zone.  In that particular case it's a little easier since we happen to do all of the Route53 managing in one centralized place.  Still, I'm somewhat surprised that having better contextual information in order to automatically provide the relevant credential hasn't come up before.  Going forward with the 2.0 SDK this is something I'd love to see a general solution for across all services not just S3.

millems:
I was discussing this issue recently: we expect customers to maintain knowledge of their resources and the policies attached to those resources, but we don't provide a programmatic way of being able to determine what resources a particular request will affect.

This makes it really difficult for customers to decide on credentials to use for a particular request without involving a person's brain power in the mix. When things get really complicated, even that starts to break down.

I don't have a better solution right now than what was suggested, unfortunately.

zarfide:
> This makes it really difficult for customers to decide on credentials to use for a particular request without involving a person's brain power in the mix. 

Our solution to this has been to attempt to enforce the paradigm that all objects in a given bucket should be owned by the same account entity as the bucket.  Given this assumption, all clients need is a mapping of bucket->account and a naming paradigm for which role to use.  And of course hooks in the API to access the logic :-)

zarfide:
Fun fact I discovered today.  While the `beforeExecution` method was added to`RequestHandler2` in `1.11.106`, it does't actually get called from `AmazonS3` requests in this version.  (It does work as expected in the latest `1.11.167`.)

millems:
Yeah, we missed it in the S3 client (because it is hand-managed, unlike the other clients) when we originally created the feature. Sorry about that confusion.

PR #1290, opened 2017-09-05
Adding credentials while creating SQS object

sakky11:
Adding credentials while creating SQS object


sakky11:
In the given example, there were no credentials provided while creating the SQS object,Adding that.

shorea:
Nice catch! Looks good to me, merging.

PR #1305, opened 2017-09-18
1.10.75.x

rsaraswati:


kiiadi:
The clock-skew fix is already in `master`. This was a back-port of the fix to the 1.10 version of the SDK. What are you trying to do exactly?

PR #1313, opened 2017-09-23
Updating S3AbortableInputStream#close abort delegation

jamesridgway:
Updating S3AbortableInputStream#close to delegate to S3ObjectInputStream#abort if there is any data remaining in the stream.
    
This is now consistent with the current documentation, and also ensurers that super.abort() is being triggered (which was being done by S3AbortableInputStream#abort and not S3AbortableInputStream#close). Therefore the abort behaviour via abort and close are now consistent.


shorea:
@jamesridgway Can you describe why the current behavior is not desirable?

dagnir:
I believe the intention with this is that when `close()` is called on a stream that hasn't been completely drained, we don't want to treat it as an abort, which is why we don't delegate to abort here despite them looking very similar atm.

shorea:
Closing this, we just pushed a big fix for the abort behavior.

Issue #1321, opened 2017-09-29
Fix downloadDirectory method in TransferManager


jperez-dx:
The current implementation of the `downloadDirectory` method creates the whole directory structure based on the key of the objects and not only their relative path to the key prefix.

For example, if the `downloadDirectory` method is called with the key prefix `sample-dir/other-dir/` and `/Users/jperez/Desktop/dest-dir/` as the destination dir, then all the S3 files with that key prefix in the given bucket would be placed in `/Users/jperez/Desktop/dest-dir/sample-dir/other-dir/`, when they should be placed under `/Users/jperez/Desktop/dest-dir/`. Files should be located in the root of the destination dir.

Maybe this can be fixed by replacing this line...
https://github.com/aws/aws-sdk-java/blob/7e064a48bee151ef222e0edea6157c1e65c7904a/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/transfer/TransferManager.java#L1271
...with something like
```java
File f = new File(destinationDirectory, summary.getKey().lstrip(keyPrefix));
```

I don't know if this is implemented this way on purpose, but it makes much more sense to do it this other way.

millems:
Unfortunately we can't change this behavior because it would break backwards compatibility with customers that currently expect the previous behavior, but I agree that your approach is more intuitive. 

We have an opportunity to break backwards compatibility a bit with our next major version of the Java SDK (v2.0). I've linked to this issue in our 2.0 repository to make sure we don't miss this suggestion when we're making the updates.

Thanks for the report!

kevin860:
@millems understood if this is more of a v2.0 type change, but if you're interested I ran into this as well today and created a pull request that won't break backwards compatibility.

alecswan:
@millems, any chance merging @kevin860 backward-compatible change into this repo?

zoewangg:
Hi all, just a quick update, this feature has been implemented in the transfer manager in Java SDK v2. Note that it's in Developer Preview. You can check out sample code here https://github.com/aws/aws-sdk-java-v2/tree/master/services-custom/s3-transfer-manager. 

As always, we welcome feedback, bug reports and feature requests on the https://github.com/aws/aws-sdk-java-v2 repo!:)

PR #1322, opened 2017-09-29
Add 'stripKeyPrefix' option to TransferManager's downloadDirectory method

kevin860:
Added an option to strip key prefixes to change downloadDirectory method behaviour without breaking backwards compatibility.

See #1321

### DIFF
--- a/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/transfer/TransferManager.java
+++ b/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/transfer/TransferManager.java
@@ -1287,7 +1287,7 @@ public class TransferManager {
 
         for ( S3ObjectSummary summary : objectSummaries ) {
             // TODO: non-standard delimiters
-            File f = new File(destinationDirectory, summary.getKey());
+            File f = new File(destinationDirectory, stripKeyPrefix ? summary.getKey().substring(keyPrefix.length()) : summary.getKey());
             File parentFile = f.getParentFile();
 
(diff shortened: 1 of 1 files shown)


aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/transfer/TransferManager.java
```diff
@@ -1178,17 +1192,22 @@ public MultipleFileDownload downloadDirectory(String bucketName, String keyPrefi
...
      * @param filter
      *           If set, applies the filter to determine which keys to include
      *           in the download request. (default is include all).
+     * @param stripKeyPrefix
+     *           If set, strips the {@code keyPrefix} from the directory structure
+     *           of the downloaded files.
      */
     public MultipleFileDownload downloadDirectory(String bucketName, String keyPrefix, File destinationDirectory,
-                                                  boolean resumeOnRetry, KeyFilter filter) {
+                                                  boolean resumeOnRetry, KeyFilter filter, boolean stripKeyPrefix) {
```
jperez-dx: You never use this flag to strip the key prefix from the object key 😕

kevin860: Oops that's embarrassing! That's what I get for re-doing my changes in a more pull request worthy way...

Thanks, I pushed the actual functionality.

millems:
Sorry about the delay on this. Can you share some test code for this functionality? We'd like to import it to our test suite before we merge this change.

PR #1340, opened 2017-10-13
Added log message to AmazonHttpClient.getNonNullResponseHandler

aharakh:
When instance of `AmazonHttpClient` is created there should be set `errorResponseHandler` of `requestExecutionBuilder`. Otherwise `nonNullResposeHandler` will be used, which, in fact, does nothing but returns null instead of exception object. Then it goes to `handleErrorResponse` method and tries to do the following ` exception.fillInStackTrace();`, which means  `NullPointerException`. As a result we have `NullPointerException` instead of any information about request's response. My suggestion is to at least log warning with pointing out the root cause of this problem: there's no custom error handler set.

### DIFF
--- a/aws-java-sdk-core/src/main/java/com/amazonaws/http/AmazonHttpClient.java
+++ b/aws-java-sdk-core/src/main/java/com/amazonaws/http/AmazonHttpClient.java
@@ -529,6 +529,8 @@ public class AmazonHttpClient {
 
                 @Override
                 public T handle(HttpResponse response) throws Exception {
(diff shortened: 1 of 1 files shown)


shorea:
@aharakh apologies for the delay, are you using AmazonHttpClient directly?

shorea:
@aharakh Let us know how you are using AmazonHttpClient so we can make the most appropriate fix.

shorea:
Closing pull request for now. If you have some time to explain your use case we can evaluate this further.

PR #1342, opened 2017-10-16
Use upper camel case conversion for JSON element names (lambda)

### LINKED ISSUE #1344: LambdaInvokerFactory doesn't work with the event sources that AWS Lambda natively accepts
For instance, when an input data based on the following POJO is passed to a Lambda function through aws-java-sdk-lambda SDK library, the Lambda function receives an empty data instead.

```
        private String eventSource;

        /**
         * Gets the event source
         * @return event source
         */
        public String getEventSource() {
            return eventSource;
        }

        /**
         * Sets the event source
         * @param eventSource A string containing the event source
         */
        public void setEventSource(String eventSource) {
            this.eventSource = eventSource;
        }
```
(this is a snippet from https://github.com/aws/aws-lambda-java-libs/blob/master/aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/SNSEvent.java)

The raw string after the serialization (by ObjectMapper in LambdaInvokerFactory) is below.

`{"eventSource":"foobar"}`

This results in an empty input to the Lambda function even if the same POJO is used. Not 'eventSource' but 'EventSource' seems to be expected for the JSON deserialization logic in AWS Lambda service.

DaichiUeura:
This PR is to make the JSON serialization/deserialization behaviour compatible with AWS Lambda service.

For instance, when an input data based on the following POJO is passed to a Lambda function through aws-java-sdk-lambda SDK library, the Lambda function receives an empty data instead.

```
        private String eventSource;

        /**
         * Gets the event source
         * @return event source
         */
        public String getEventSource() {
            return eventSource;
        }

        /**
         * Sets the event source
         * @param eventSource A string containing the event source
         */
        public void setEventSource(String eventSource) {
            this.eventSource = eventSource;
        }
```
(this is a snippet from https://github.com/aws/aws-lambda-java-libs/blob/master/aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/SNSEvent.java)

The raw string after the serialization (by ObjectMapper in LambdaInvokerFactory) is below.

`{"eventSource":"foobar"}`

This results in an empty input to the Lambda function even if the same POJO is used. Not 'eventSource' but 'EventSource' seems to be expected for the JSON deserialization logic in AWS Lambda service. So if upper camel case conversion rule is applied to the ObjectMapper in LambdaInvokerFactory, the raw string becomes `{"EventSource":"foobar"}` and then works perfectly.

Fix #1344

### DIFF
--- a/aws-java-sdk-lambda/src/main/java/com/amazonaws/services/lambda/invoke/LambdaInvokerFactory.java
+++ b/aws-java-sdk-lambda/src/main/java/com/amazonaws/services/lambda/invoke/LambdaInvokerFactory.java
@@ -27,6 +27,7 @@ import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.DeserializationFeature;
 import com.fasterxml.jackson.databind.JavaType;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.PropertyNamingStrategy;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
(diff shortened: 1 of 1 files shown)


shorea:
Unfortunately this would be a breaking change. Do you need the invoker factory for SNS events? I thought those are generally triggered by SNS itself rather than manually invoking the function (unless your trying to write a functional test or something).

DaichiUeura:
> unless your trying to write a functional test or something
Yes, I'm trying to write end to end test.

Would it be possible to change a code around LambdaInvokerFactory to be able to pass a configuration(that controls ObjectMapper behavior)?

DaichiUeura:
At least, this pull request breaks something and also this doesn't resolve other cases such as DynamoDB Streams input case. So let me close until a good solution is found.

PR #1355, opened 2017-10-22
Support for IoT actions

MrLesh:
Similiar to [S3Actions](https://github.com/aws/aws-sdk-java/blob/master/aws-java-sdk-s3/src/main/java/com/amazonaws/auth/policy/actions/S3Actions.java) or any other supported service actions - wanted to have IoT actions that I can refer to in a code.
I'm using it to create IoT policies programmatically


shorea:
This is a generated class, let me turn on generation and see how it compares.

shorea:
Ok ran into a bit of a problem. We generate these actions based on the APIs present in the service model which doesn't always match up to the actual actions (which is mostly why we don't generate them for newer services). So a couple of the actions in your PR would not be generated by our code generator (Connect is one example). I'd personally not like to have to hand maintain this file as once it's generated I think the expectation is that it's complete which means we have to hand edit it everytime IoT does a release. Would it meet your needs if you just maintained your own version of this enum in your application with just the actions you use/need?

shorea:
Closing for now. If the suggested workaround isn't viable let us know and we can discuss further.

Issue #995, opened 2017-01-23
UploadPartRequest partSize is a required field, but not listed in comments or properly validated


landon9720:
The comment at the top of `UploadPartRequest.java` does not list `partSize` as a required field.

If `partSize` is not set, `AmazonS3Client#uploadPart` should fail. The current implementation is incorrect because it checks that `partSize` is not null, however `partSize` is a primitive long and therefore is never null.

kiiadi:
@landon9720 I've switched the underlying datatypes to their boxed equivalents and updated the validation to use those. I've also updated the comment at the top. Should be out in our next release.

kiiadi:
@landon9720 thinking more about this, I'm not sure that we can do this validation without it being backwards incompatible. Currently, if you do not supply partSize on the UploadPart the request will succeed - the part that's uploaded will be zero bytes long. Whilst this might be undesired there may be customers relying on this behaviour. I think we have to leave it as it is for now; I agree we might as well remove the validation.

kayvonr:
Would it be possible to have them boxed but still accept 0 as a value? That was forces people to explicitly set to 0 if they want, otherwise errors out if someone omits it. Does not matter for me anymore, I know now that I need it, just thinking of people using in the future.

shorea:
@kayvonr I believe that might also be considered a breaking change if customers are implicitly relying on zero being the default.

PR #1362, opened 2017-10-25
Make partSize field of UploadPartRequest required and non-primitive

### LINKED ISSUE #1360: S3 Multipart upload fails on ByteArrayInputStream passed as part content body.
I'm trying to run multipart upload and pass `byte[]` as part body by wrapping it with `ByteArrayInputStream`. Simple example:
```java
String bucket = "test-bucket";
String key = "test1";
AmazonS3 s3 = //...
Random rand = new Random();
Function<Integer, byte[]> randomBytes = (size) -> {
    byte[] bb = new byte[size];
    rand.nextBytes(bb);
    return bb;
};

String uploadId = s3
    .initiateMultipartUpload(new InitiateMultipartUploadRequest(bucket, key))
    .getUploadId();
int partSize = 5 * 1024 * 1024;
List<PartETag> list = new ArrayList<>(3);

for (int i = 1; i <= 3; i++) {
    UploadPartRequest req = new UploadPartRequest()
        .withInputStream(new ByteArrayInputStream(randomBytes.apply(partSize)))
        .withBucketName(bucket)
        .withUploadId(uploadId)
        .withKey(key)
        .withPartNumber(i);
    list.add(s3.uploadPart(req).getPartETag());
}

s3.completeMultipartUpload(new CompleteMultipartUploadRequest(bucket, key, uploadId, list));
```

However on the completion I'm getting `AmazonS3Exception: Your proposed upload is smaller than the minimum allowed object size. (Service: Amazon S3; Status Code: 400; Error Code: EntityTooSmall)`.

Here's part of logs:
```
-  http-outgoing-0 >> PUT /test-bucket/test1?uploadId=9cf8c626-6cbc-4973-a2c1-6d65f188fdf3&partNumber=2 HTTP/1.1
-  http-outgoing-0 >> x-amz-content-sha256: STREAMING-AWS4-HMAC-SHA256-PAYLOAD
-  http-outgoing-0 >> X-Amz-Date: 20171024T142139Z
-  http-outgoing-0 >> User-Agent: aws-sdk-java/1.11.213 Linux/4.10.0-37-generic Java_HotSpot(TM)_64-Bit_Server_VM/25.144-b01 java/1.8.0_144
-  http-outgoing-0 >> x-amz-decoded-content-length: 0
-  http-outgoing-0 >> amz-sdk-retry: 0/0/500
-  http-outgoing-0 >> Content-Type: application/octet-stream
-  http-outgoing-0 >> Content-Length: 86
-  http-outgoing-0 >> Connection: Keep-Alive
-  http-outgoing-0 >> Expect: 100-continue
-  http-outgoing-0 << HTTP/1.1 200 OK
-  http-outgoing-0 << Accept-Ranges: bytes
-  http-outgoing-0 << Etag: "d41d8cd98f00b204e9800998ecf8427e"
```

As you can see `Content-Length` is 86 but not 5MB.

Is there a way to upload dynamic data from memory instead of writing it to file and passing it to request ?

mentegy:
Fixes #1360

When passing `withInputStream` into `UploadPartRequest` it's easy to forgot adding `withPartSize` value. Since it's primitive, java makes it 0 and complete upload will fail at the completion stage. 

This PR makes it `Long` object and adds `rejectNull` validation

### DIFF
--- a/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/AmazonS3Client.java
+++ b/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/AmazonS3Client.java
@@ -3146,13 +3146,15 @@ public class AmazonS3Client extends AmazonWebServiceClient implements AmazonS3 {
         final String key        = uploadPartRequest.getKey();
         final String uploadId   = uploadPartRequest.getUploadId();
         final int partNumber    = uploadPartRequest.getPartNumber();
-        final long partSize     = uploadPartRequest.getPartSize();
+        final Long partSize     = uploadPartRequest.getPartSize();
         rejectNull(bucketName,
             "The bucket name parameter must be specified when uploading a part");
         rejectNull(key,
             "The key parameter must be specified when uploading a part");
         rejectNull(uploadId,
             "The upload ID parameter must be specified when uploading a part");
+        rejectNull(partSize,
+                "The part size parameter must be specified when uploading a part");
         Request<UploadPartRequest> request = createRequest(bucketName, key, uploadPartRequest, HttpMethodName.PUT);
         request.addParameter("uploadId", uploadId);
         request.addParameter("partNumber", Integer.toString(partNumber));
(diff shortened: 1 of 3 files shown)


mentegy:
UPD. When user desires to set 0 by not setting partSize at all, he should avoid implicitness and set  `withPartSize(0)` instead

shorea:
Unfortunately this would be a breaking change for customers. We've discussed adding validation for that field but that too might be considered a breaking change, see https://github.com/aws/aws-sdk-java/issues/995. For now we'll just have to live with this behavioral quirk, in V2 of the Java SDK the S3 client is generated and uses boxed types for everything though.

PR #1367, opened 2017-11-01
Use 8k buffers when downloading into file

ksperling:
Writes <= 8k use a stack-allocated buffer whereas larger writes will malloc() / free() a fresh buffer for each write. Since there doesn't appear to be any particular reason to use a 10k buffer, it makes sense to align with the "preferred" buffer size of the JVM of 8k. See http://hg.openjdk.java.net/jdk8u/jdk8u/jdk/file/601a2304ff4a/src/share/native/java/io/io_util.c#l58

### DIFF
--- a/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/internal/ServiceUtils.java
+++ b/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/internal/ServiceUtils.java
@@ -68,6 +68,9 @@ public class ServiceUtils {
     public static final boolean OVERWRITE_MODE = false;
 
     private static final SkipMd5CheckStrategy skipMd5CheckStrategy = SkipMd5CheckStrategy.INSTANCE;
+    
+    // Preferred buffer size of JVM's native FileOutputStream.write() implementation.
+    private static final int DEFAULT_BUFFER_SIZE = 8192;
 
     @Deprecated
     protected static final DateUtils dateUtils = new DateUtils();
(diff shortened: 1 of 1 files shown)


shorea:
Seems pretty reasonable. Do you have any data on the performance gains for this change?

aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/internal/ServiceUtils.java
```diff
@@ -302,7 +302,7 @@ public static void downloadToFile(S3Object s3Object,
             }
             outputStream = new BufferedOutputStream(new FileOutputStream(
                     dstfile, appendData));
-            byte[] buffer = new byte[1024*10];
```
kiiadi: `8192` isn't _that_ difficult to figure out but it might be worth putting the magic number into a well named constant variable.

ksperling:
@shorea Hard to say. It may actually be slower... in a small synthetic benchmark I put together just now the cost of having 25% more separate IO calls seems to outweigh the cost of the malloc/free, and quadrupling the buffer size to 40k instead approximately doubles throughput.

```
IoBenchmark.writeByteArray    8192  thrpt   20  152993.686 ± 1789.748  ops/s
IoBenchmark.writeByteArray   10240  thrpt   20  159904.641 ± 1396.147  ops/s
IoBenchmark.writeByteArray   40960  thrpt   20  327525.351 ± 1992.442  ops/s
```

https://gist.github.com/ksperling/d9f37f1285a0722ec41e13c13299584e

In practice the difference is probably unnoticeable compared to the cost of real network and disk IO.

millems:
Thanks for taking time to test this out! Did you still want to move forward and have us look at this change or are things okay as they are?

ksperling:
Hm your call really. Using a 10k buffer without any particular justification still seems very odd to me given that 8k is something of a "standard". Apart from the native code I linked to it is also the default buffer size of `BufferedOutputStream` and `BufferedInputStream` as well as being a multiple of the common page size of 4k. So if this was new code being written now I would say to go with the conventional 8k unless there was a good justification for deviating from it.

dagnir:
I agree if this was being written today, 8 KB buffers should be what we use. For now since there's no compelling data showing whether 10 KB causes a measurable performance dip, I'm going to close this PR. If we can get more data, we're more than happy to revisit this again!

PR #1380, opened 2017-11-10
POM: Fix SCM meta-data

sschuberth:
The URL should be browsable and not end in ".git". Add missing (anonymous)
connection and (authenticated) developerConnection tags.


dagnir:
LGTM

PR #1412, opened 2017-12-08
Add us-east-1 to S3 Regions

### LINKED ISSUE #1411: US-East-1 Missing from S3 Regions (was hard to find reference that US_Standard should be used)
http://docs.aws.amazon.com/AmazonS3/latest/dev/Introduction.html indicates that S3 does run in us-east-1, but you won't find an enum value here: https://github.com/aws/aws-sdk-java/blob/master/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/model/Region.java#L69

This page indicates that us-east-1 is US_Standard: http://docs.aws.amazon.com/general/latest/gr/rande.html#s3_region

Might be nice to mention this in the docs somehow.

timcharper:
Fixes #1411

### DIFF
--- a/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/model/Region.java
+++ b/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/model/Region.java
@@ -57,6 +57,18 @@ public enum Region {
     US_Standard((String[])null),
 
     /**
+     * The US-East-1 (N. Virginia) Region. This region
+     * uses Amazon S3 servers located in Ohio.
+     * <p>
+     * When using buckets in this region, set the client
+     * endpoint to <code>s3.us-east-1.amazonaws.com</code> on all requests to these buckets
+     * to reduce any latency experienced after the first hour of
+     * creating a bucket in this region.
+     * </p>
+     */
+    US_East("us-east-1"),
+    
+    /**
      * The US-East-2 (Ohio) Region. This region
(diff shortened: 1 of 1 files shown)


varunnvs92:
There is no official us-east-1 endpoint for s3. The global endpoint (s3.amazonaws.com) is considered to the us-east-1 endpoint. That region is represented by [US_Standard](https://github.com/timcharper/aws-sdk-java/blob/8640e2d6ee1d3dbee1215a252de5b3241bf3c606/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/model/Region.java#L57) enum value.

timcharper:
I was wondering if that might be the case but [the S3 Introduction docs](http://docs.aws.amazon.com/AmazonS3/latest/dev/Introduction.html) persuaded me otherwise.

Okay, then a note in the docs might be nice for future people that stumble here?

varunnvs92:
If you check the note [here](http://docs.aws.amazon.com/general/latest/gr/rande.html#s3_region), you will find "Amazon S3 renamed the US Standard Region to the US East (N. Virginia) Region". When s3 is launched, us-east-1 is represented by US standard which is what the SDK represented in the enum value. Even though S3 changed their documentation, we can't use a new value in our enums as it is a breaking change for existing customers.

timcharper:
K, closing; updating the issues link with a potential doc improvement.

PR #1437, opened 2018-01-05
Moved the assertion verification out of the setEnabledProtocols method.

oscarlvp:
If there is a bug in prepareSocket that prevents setEnabledProtocols from being called, the covering tests never fail as no assertion is verified. The extreme case is that you can remove the entire code of prepareSocket and no test will complain about it. This pull request contains a small change to improve the tests in order to catch that kind of error.


aws-java-sdk-core/src/test/java/com/amazonaws/http/conn/ssl/TestSSLSocket.java
```diff
@@ -15,13 +15,45 @@
...
+import java.util.Arrays;
 
 import javax.net.ssl.HandshakeCompletedListener;
 import javax.net.ssl.SSLSession;
 import javax.net.ssl.SSLSocket;
 
 class TestSSLSocket extends SSLSocket {
 
+    @Override
+    public void setEnabledProtocols(String[] protocols) {
```
shorea: I agree with the rationale behind this change, I think it improves the test quality. I'm thinking that it would be better to have TestSSLSocket just capture the protocols passed to setEnabledProtocols and then the test code can query that and make assertions based on that. That gives you a bit more flexibility on what you can assert on since it moves the assertion out of this mock object and into the respective tests.

So something like

TestSSLSocket socket = ...
f.prepareSocket(socket)
Assert.arrayEquals(socket.getCapturedProtocols(), expectedProtocols);

oscarlvp:
I made the changes to match the approach proposed by @shorea which is cleaner than my initial attempt.

shorea:
Looks great! Can you confirm you are submitting under the Apache 2.0 license?

oscarlvp:
Yes, sure. Should I do something specific?

shorea:
Just written confirmation on this PR via a comment will suffice.

shorea:
Thank you for the contribution @oscarlvp! This will be included in the next release of the SDK.

oscarlvp:
By the way, the issue behind this fix was discovered using tools developed by the [STAMP project](https://www.stamp-project.eu/main/), specifically [this one](https://github.com/STAMP-project/pitest-descartes). We may have more results in the near future. Would you be interested in our findings? We can keep collaborating in this direction.

PR #1446, opened 2018-01-11
Adds toItem function to InternalUtils

hankwallace:
Adds the `toItem` function to convert a map of item information to an `Item`. There was already a `toItemList` function in there, but not one for a single item.

### DIFF
--- a/aws-java-sdk-dynamodb/src/main/java/com/amazonaws/services/dynamodbv2/document/internal/InternalUtils.java
+++ b/aws-java-sdk-dynamodb/src/main/java/com/amazonaws/services/dynamodbv2/document/internal/InternalUtils.java
@@ -49,6 +49,16 @@ public enum InternalUtils {
     ;
 
     /**
+     * Returns an <code>Item</code> given the low level item information;
+     * or null if the input is null;
+     */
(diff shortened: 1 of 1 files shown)


spfink:
Hmm, this class is internal and specifically noted that it is not meant for general use and may change at any time. We typically don't have strong backwards compatibility guarantees for this.

Let me do some digging to see how widely used this class is. Either need to make this class not in our internal folder or might need a new class for this method.

hankwallace:
Some of the utilities in here are very useful and it would be great to have them in a different folder. ex: `toAttributeValues`, `toSimpleMapValue`

shorea:
@hankwallace we've decided to expose this class publicly. We have deprecated the existing InternalUtils and introduced a new ItemUtils class with the same functionality. I'll go ahead and add this method to that class and get it out for our next release. 

I'd like to accept your pull request as well, can you please confirm you are submitting under the Apache 2.0 license?

hankwallace:
Sounds good. Yes, I submitting under the Apache 2.0 license.

shorea:
Great! I've gone ahead and staged this for our next release. Just FYI the new method will be in com.amazonaws.services.dynamodbv2.document.ItemUtils. Thanks for the contribution!

PR #1489, opened 2018-02-26
Update jackson to use a later stable version (2.7.9) and branch (2.7)

derekstraka:


shorea:
We've discussed this internally. Since 2.7 also has security issues we have decided not to upgrade the SDK dependency. Consumers of the SDK are free to override the version of Jackson used to a newer version as they are ABI compatible.

PR #1508, opened 2018-03-15
Cognito connectors

mataide:
I have created Cognito Connectors, just like it works for Android.
It helps a lot when you use start to use Cognito.


millems:
Sorry this pull request was ignored for so long. We're not currently taking non-trivial pull requests for 1.11.x, because most development effort is focused on 2.x.

Issue #1502, opened 2018-03-12
S3 Transfer Manager bytesTransferred not being updated


stevorobs3:
I am using the S3 Transfer Manager within a Scala project. It seems like the bytesTransferred is not being updated. Please see below sample code:

 val amazonTransferManager : TransferManager = ???
 val copy = amazonTransferManager.copy(sourceBucket, sourceKey, destinationBucket, destinationKey)
 while (!copy.isDone) {
   Thread.sleep(100)
   println(s"Transfer progress ${copy.getProgress.getBytesTransferred} / ${copy.getProgress.getTotalBytesToTransfer}")
 }

The `getTotalBytesToTransfer` always returns the correct value but the `getBytesTransferred` always returns 0 even when the `TransferState` is Completed.

Is this expected behaviour or is there something I can do differently to ensure the `getBytesTransferred` method returns the expected result?

Thanks,

Steve

millems:
It's unfortunate, but for `copy` calls, S3 doesn't return any progress indicator that can be used to supply you with the bytes transferred. This is because the copy is entirely service-side. We get the `totalBytesToTransfer` from the object metadata.

The `byteTransferred` only works for uploads and downloads.

stevorobs3:
Ok thanks very much for clarifying this. Would a pr be welcome that documents this in code? Also, could we set the bytes transferred = totalbytestotransfer when the progress state changes to complete?

millems:
A PR is always welcome, but we're willing to fix it ourselves. Let me know which you prefer.

I agree the documentation isn't clear and updating byte transferred once we finish the transfer makes sense.

stevorobs3:
Great I will have a look at making that change now. Do you have any developer guidelines that I can read? I can't find any contributor guidelines.

stevorobs3:
I have added PR #1510 in order to close this

millems:
Merged the PR! It should be released to maven today or Monday.

JoseMFranco:
Hi guys, so I'm wondering if the fix is working in the Java SDK. I used it with maven v1.11.340
According to the dates that fix should be there, right?
But I'm still getting 0 when I used the copy call and the bytesTransferred() method.

I would really appreciate your help. Thanks.

PR #1510, opened 2018-03-15
S3 copy update the bytesTransferred on Copy completion

stevorobs3:
This is to fix issue #1502 by improving the documentation and updating the bytesTransferred on Copy completion

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,6 @@
 target/
 *settings.xml
+
+#intellij idea generated files
+.idea/
+*.iml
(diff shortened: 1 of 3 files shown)


millems:
Thanks for your contribution! Can you confirm you're fine submitting it under the Apache License, Version 2.0 (http://aws.amazon.com/apache2.0)?

stevorobs3:
Yes I'm more than happy to submit under that license.

As a side note: I wonder if there is a way to automate that question using travis?

millems:
Great question to ask! In our V2 repository, we automate that question with a PR template. We haven't back-filled that to the V1 repository quite yet.

PR #1526, opened 2018-03-29
Fix incompatible JAXB implementation detection in Base64

VladRassokhin:


VladRassokhin:
@millems can you please take a look? And also #1527

PR #1535, opened 2018-04-03
Bump Jackson to 2.9.5 to address security vulnerability CVE-2018-7489

somechris:
Jackson <=2.9.4 is affected by CVE-2018-7489, which allows
unauthenticated remote code execution.

http://cve.circl.lu/cve/CVE-2018-7489

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -176,8 +176,8 @@
   </scm>
   <properties>
       <awsjavasdk.version>${project.version}</awsjavasdk.version>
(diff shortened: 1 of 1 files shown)


spfink:
Removing Java 6 is not an option at this time but it is something that we want to look into.

So unfortunately we can't pull this change in. Feel free to use any version of Jackson within your project though as it should be code compatible.

somechris:
Gotcha. Thanks!

PR #1555, opened 2018-04-19
Fixed Route53 HostedZone Id on ListHostedZonesByName response

dcrn:
For the Route53 `ListHostedZonesByName` response, the prefix from HostedZone's Id field wasn't being stripped as it should be. This caused HostedZone objects returned to not work with other requests.

before:
```java
System.out.println(route53.listHostedZones().getHostedZones().get(0).getId());
// output: ABCDEF12345678
System.out.println(route53.listHostedZonesByName().getHostedZones().get(0).getId());
// output: /hostedzone/ABCDEF12345678
```

after:
```java
System.out.println(route53.listHostedZones().getHostedZones().get(0).getId());
// output: ABCDEF12345678
System.out.println(route53.listHostedZonesByName().getHostedZones().get(0).getId());
// output: ABCDEF12345678
```

### DIFF
--- a/aws-java-sdk-route53/src/main/java/com/amazonaws/services/route53/internal/Route53IdRequestHandler.java
+++ b/aws-java-sdk-route53/src/main/java/com/amazonaws/services/route53/internal/Route53IdRequestHandler.java
@@ -28,6 +28,7 @@ import com.amazonaws.services.route53.model.GetHostedZoneResult;
 import com.amazonaws.services.route53.model.GetReusableDelegationSetResult;
 import com.amazonaws.services.route53.model.HostedZone;
 import com.amazonaws.services.route53.model.ListHostedZonesResult;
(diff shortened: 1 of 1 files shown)


dcrn:
Yes, this PR is submitted under Apache 2.0.

varunnvs92:
Thanks for confirming. This change will be included in next release.

varunnvs92:
@dcrn 
Hi, This change broke customers who were relying on the old ID format. As this API is 3 years old and most customers using the API would be relying on the old format, we will be reverting the change to not break them. Please change your application accordingly. Sorry for the inconvenience. Thank you.

dcrn:
@varunnvs92 
In that case maybe we should remove `/hostedzone/` from the ID (if present) on usage in other Route53 methods?

PR #1589, opened 2018-05-15
ignore .idea directory created by intellij

fs111:


spfink:
Thanks for the contribution!

Issue #1436, opened 2018-01-05
Lambda: Provide access to ObjectMapper in LambdaInvokerFactory


gerrit-hntschl:
Hi,

I need to set the date format for the `ObjectMapper` and now have to resort to reflection.
Please make the static `MAPPER` accessible or allow to set your own instance.

Best regards,
Gerrit

kiiadi:
Can you provide an example of where you need this? I'm not sure I fully understand the context.

gerrit-hntschl:
Thanks for the response. 

What I am doing is this:
```
class RequestParams{
 private Date startDate;

 public Date getStartDate(){
   return startDate;
 }
}
```
The lambda function expects dates to be formatted as a string according to ISO8601.

To invoke the lambda I use this:

```
public interface SomeLambda {
    @LambdaFunction
    Response doSomething(RequestParams request);
}

LambdaInvokerFactory.builder()
                .lambdaClient(lambda)
                .build(SomeLambda.class).doSomething(new RequestParams(someStartDate));

```
Under the hood the LambdaInvokerFactory uses the static field `ObjectMapper MAPPER` to turn my `RequestParams` into JSON.

The problem is I cannot access that mapper to tell it how to serialize the date. Currently I have to hack my way around that like this:
```
            Field mapperField = LambdaInvokerFactory.class.getDeclaredField("MAPPER");
            mapperField.setAccessible(true);

            Object mapperVal = mapperField.get(null);
            if (!(mapperVal instanceof ObjectMapper)) {
                throw new IllegalStateException("Mapper not an objectmapper: " + mapperVal.getClass());
            }
            ObjectMapper objectMapper = (ObjectMapper) mapperVal;

            objectMapper.setDateFormat(new ISO8601DateFormat());
```

For my use-case having `.objectMapper(...`-method on the `LambdaInvokerFactory.Builder` would be fine.

shorea:
Would the JsonFormat annotation work?

http://www.baeldung.com/jackson-jsonformat

shorea:
Closing this issue, if the mentioned annotation doesn't work out feel free to reopen and we can explore other options.

etwillbefine:
In our use case the `@JsonFormat` annotation is not very suitable. We have defined one object mapper for all our applications und dtos. I dont want to go through each `Date` property and add an annotation just because there is no easy way to override the static mapper which in any case should be configureable.

avirtuos:
[...]

PR #1591, opened 2018-05-17
feat: make object mapper for lambda configureable

etwillbefine:
1. Making the `ObjectMapper` configureable 
2. Returning result of `.hashCode` to avoid `LambdaSerializationException` as it prevents us from defining a lambda function interface as bean in a spring boot application. See also: https://stackoverflow.com/a/49071474/3703307

I found no unit tests therefore I'm quite unsure how to test the changes or if its even required?!

I hope we can get this merged soon because its very important for us to use our own `ObjectMapper` and not defining additional annotations  

closes https://github.com/aws/aws-sdk-java/issues/1436

### DIFF
--- a/aws-java-sdk-lambda/src/main/java/com/amazonaws/services/lambda/invoke/LambdaInvokerFactory.java
+++ b/aws-java-sdk-lambda/src/main/java/com/amazonaws/services/lambda/invoke/LambdaInvokerFactory.java
@@ -70,7 +70,7 @@ import java.nio.ByteBuffer;
  */
 public final class LambdaInvokerFactory {
 
-    private static final ObjectMapper MAPPER = new ObjectMapper()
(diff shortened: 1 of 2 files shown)


shorea:
@etwillbefine can you confirm that you are submitting under the Apache 2.0 license?

etwillbefine:
Yes

shorea:
Merging, thanks for the contribution! I made a small change to require the ObjectMapper passed in to have the FAIL_ON_UNKNOWN_PROPERTIES disabled to ensure forward compatibility if the lambda function returns a new field the client doesn't have modeled.

etwillbefine:
Thanks for merging. I guess a new release is coming soon? :)

shorea:
@etwillbefine yes it will be out very shortly.

PR #1595, opened 2018-05-23
Add descriptive names for AWS regions

### LINKED ISSUE #1570: Regions should have proper, easily understandable 'Region Names'
The Enum `Regions` right now have just the code. It would be better to also include a proper, easy to understand name. For ex: `us-west-1` should also have the proper name `US West (N. California)`. This may sound trivial but I think this might come in handy in many situations for third party softwares and/or even AWS SDKs.

BigChief45:
Closes #1570

### DIFF
--- a/aws-java-sdk-core/src/main/java/com/amazonaws/regions/Regions.java
+++ b/aws-java-sdk-core/src/main/java/com/amazonaws/regions/Regions.java
@@ -25,24 +25,24 @@ import com.amazonaws.util.EC2MetadataUtils;
  */
 public enum Regions {
 
-    GovCloud("us-gov-west-1"),
-    US_EAST_1("us-east-1"),
-    US_EAST_2("us-east-2"),
-    US_WEST_1("us-west-1"),
-    US_WEST_2("us-west-2"),
-    EU_WEST_1("eu-west-1"),
-    EU_WEST_2("eu-west-2"),
-    EU_WEST_3("eu-west-3"),
-    EU_CENTRAL_1("eu-central-1"),
-    AP_SOUTH_1("ap-south-1"),
-    AP_SOUTHEAST_1("ap-southeast-1"),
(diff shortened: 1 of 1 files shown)


zoewangg:
@BigChief45 Thanks for sending out PR! Can you confirm that you are submitting under the Apache 2.0 license?

BigChief45:
@zoewangg 

I confirm that I am submitting this pull request under the Apache 2.0 license.

By the way do you know the build fails for `oraclejdk8`? I tried looking at the error logs but can't seem to find the reason...

aws-java-sdk-core/src/main/java/com/amazonaws/regions/Regions.java
```diff
@@ -62,6 +64,13 @@ private Regions(String name) {
     public String getName() {
         return name;
     }
+    
+    /**
+     * Descriptive readable name for this region.
+     */
+    public String getReadableName() {
```
zoewangg: Can we rename it to `getDescription` and rename the member variable to `description` so that it matches with what we have in endpoints.json?

https://github.com/aws/aws-sdk-java/blob/6927b2eda56da6b0737aff14d1d02775011c4547/aws-java-sdk-core/src/main/resources/com/amazonaws/partitions/endpoints.json#L14

zoewangg:
That might be a transient issue. From the logs, `SDKDefaultRetryConditionTest.shouldNotRetryBad4xxErrorCodeAse:` test failed, but don't think that's related to your changes.

aws-java-sdk-core/src/main/java/com/amazonaws/regions/Regions.java
BigChief45: Sure thing!

BigChief45:
@zoewangg PR updated with requested changes.

PR #1596, opened 2018-05-24
Attempt to parse versionID from any valid URI object

umeat:
The URI class supports query string parameters, so versionId could potentially be extracted from any valid URI object.

I would like to create S3 URIs with version ID, formatted like `s3://bucket/object?versionId=version`. It is handy for storing object references without having to use the entire URL.


umeat:
Does anyone have any thoughts on this?

dagnir:
I think using the `s3://` scheme that query parameters are supported. We simply parse it as `s3://<BUCKET>/<KEY>`, so the key for `s3://bucket/object?versionId=version` would be `object?versionId=version`. This is the same behavior in the CLI/Boto3 as well.

dagnir:
Closing this as it's a behavior change and I don't think we can do this.

Issue #1632, opened 2018-06-15
Allow usage of BouncyCastle FIPS library


presto9292:
Using client-side encryption with CryptoMode.AuthenticatedEncryption fails with UnsupportedOperationException when the BouncyCastle library being used is the FIPS version
 
`org.bouncycastle.jcajce.provider.BouncyCastleFipsProvider`

Is solving this just a matter of relaxing the BouncyCastle check in CryptoRuntime to allow for the FIPS provider as well ?

```
AmazonS3EncryptionClientBuilder.standard()
.withCryptoConfiguration(newCryptoConfiguration(CryptoMode.AuthenticatedEncryption)
.withCryptoProvider(new BouncyCastleFipsProvider()))

java.lang.UnsupportedOperationException: The Bouncy castle library jar is required on the classpath to enable authenticated encryption
	at com.amazonaws.services.s3.model.CryptoConfiguration.check(CryptoConfiguration.java:241)
```

varunnvs92:
What is the difference between normal bouncy castle vs Fips version? Is it recommended to use FIPS?

varunnvs92:
This code might be relavant to your use case
https://github.com/aws/aws-sdk-java/blob/4734de6fb0f80fe5768a6587aad3b9d0eaec388f/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/internal/crypto/S3CryptoModuleBase.java#L529-L530
https://github.com/aws/aws-sdk-java/blob/4734de6fb0f80fe5768a6587aad3b9d0eaec388f/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/internal/crypto/ContentCryptoMaterial.java#L900-L915


**I would say set this key-value pair using [EncryptionMaterials#addDescription](https://github.com/aws/aws-sdk-java/blob/4734de6fb0f80fe5768a6587aad3b9d0eaec388f/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/model/EncryptionMaterials.java#L103-L106). Let us know if it worked.**

presto9292:
FIPS-140 is a US federal standard that specifies requirements for cryptography.
BouncyCastle is providing an alternative JCE implementation jar file that has been certified by a testing lab. Software for the US government is required to use the FIPS JCE version rather then the plain one. The two JCE providers can not co-exist on the classpath.
So it's either 
org.bouncycastle.jcajce.provider.BouncyCastleFipsProvider
or
[...]

PR #1633, opened 2018-06-16
Allow the use of CryptoMode.AuthenticatedEncryption with the FIPS-certified BouncyCastle

fernomac:
Hi friends!

Here's a proposed fix for #1632, as I suspect we're going to run into this at some point as well. This gets things working with the BC-FIPS library in non-approved mode; approved mode also requires some changes to the way CEKs are generated, which I'll take a look at if you're okay merging this.

The most-relevant change is to `ContentCryptoScheme`: prioritizing the user-specified security provider (in this case, `BouncyCastleFipsProvider`) over the default one called for by the chosen CryptoMode. This is, potentially, a breaking change if someone is specifying a provider but relying on the fact that it gets ignored. That _seems_ like an acceptable risk to me, but I could live with having to opt in somehow if you want to be extra extra safe.

The second change, to `CryptoConfiguration`, simply skips the BouncyCastle-related checks if the user has specified a provider to use. `check()` is called directly from `set/withCryptoMode`, so you have to call `set/withSecurityProvider` first if you don't want the checks. That's a bit surprising, but I didn't want to attempt a more serious refactor to address it.

As a bonus: Java 9 added a public `InputStream.readAllBytes()` method, breaking compilation of `S3AbortableInputStream` (which tries to define a private method of the same name). The fix is backwards compatible, so I went ahead and made it. If you would rather not bother, it's in a separate commit so it's easy to drop.

I'll preemptively say that I'm happy to contribute these changes under the terms of the Apache 2.0 license if no CLA paperwork is required. :)

### DIFF
--- a/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/internal/crypto/AesGcm.java
+++ b/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/internal/crypto/AesGcm.java
@@ -40,7 +40,7 @@ class AesGcm extends ContentCryptoScheme {
      * Currently only Bouncy Castle can support the AES/GCM cipher in
      * Java 6 without having to use the AEAD API in Java 7+.
      */
-    @Override String getSpecificCipherProvider() { return "BC"; }
+    @Override String getPreferredCipherProvider() { return "BC"; }
 
     @Override
     CipherLite createAuxillaryCipher(SecretKey cek, byte[] ivOrig,
@@ -49,7 +49,7 @@ class AesGcm extends ContentCryptoScheme {
             NoSuchProviderException, NoSuchPaddingException,
             InvalidAlgorithmParameterException {
         byte[] iv = AES_CTR.adjustIV(ivOrig, startingBytePos);
-        return AES_CTR.createCipherLite(cek, iv, cipherMode, securityProvider);
+        return AES_CTR.createCipherLite(cek, iv, cipherMode, securityProvider, false);
     }
 
     @Override
--- a/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/internal/crypto/CipherLite.java
+++ b/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/internal/crypto/CipherLite.java
@@ -86,7 +86,7 @@ class CipherLite {
      */
     CipherLite recreate() {
         return scheme.createCipherLite(secreteKey, cipher.getIV(),
-                this.cipherMode, cipher.getProvider());
+                this.cipherMode, cipher.getProvider(), true);
     }
 
     /**
@@ -95,7 +95,7 @@ class CipherLite {
      */
     CipherLite createUsingIV(byte[] iv) {
         return scheme.createCipherLite(secreteKey, iv, this.cipherMode,
-                cipher.getProvider());
+                cipher.getProvider(), true);
     }
 
     /**
@@ -128,7 +128,7 @@ class CipherLite {
         else
             throw new UnsupportedOperationException();
         return scheme.createCipherLite(secreteKey, cipher.getIV(),
-                inversedMode, cipher.getProvider());
+                inversedMode, cipher.getProvider(), true);
     }
 
     /**
@@ -162,7 +162,7 @@ class CipherLite {
      *                if this cipher is in decryption mode, and (un)padding has
      *                been requested, but the decrypted data is not bounded by
      *                the appropriate padding bytes
-     * @exception BadTagException
+     * @exception javax.crypto.AEADBadTagException
      *                if this cipher is decrypting in an AEAD mode (such as
      *                GCM/CCM), and the received authentication tag does not
      *                match the calculated value
--- a/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/internal/crypto/ContentCryptoScheme.java
+++ b/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/internal/crypto/ContentCryptoScheme.java
@@ -177,10 +177,10 @@ abstract class ContentCryptoScheme {
         String specificProvider = getSpecificCipherProvider();
         Cipher cipher;
         try {
-            if (specificProvider != null) { // use the specific provider if defined
-                    cipher = Cipher.getInstance(getCipherAlgorithm(), specificProvider);
-            } else if (securityProvider != null) { // use the one optionally specified in the input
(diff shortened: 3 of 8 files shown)


aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/internal/crypto/ContentCryptoScheme.java
```diff
@@ -177,10 +177,10 @@ CipherLite createCipherLite(SecretKey cek, byte[] iv, int cipherMode,
         String specificProvider = getSpecificCipherProvider();
         Cipher cipher;
         try {
-            if (specificProvider != null) { // use the specific provider if defined
-                    cipher = Cipher.getInstance(getCipherAlgorithm(), specificProvider);
-            } else if (securityProvider != null) { // use the one optionally specified in the input
+            if (securityProvider != null) { // use the one optionally specified in the input
```
dagnir: Hmm I'm a little apprehensive this change in precedence since `securityProvider` is clearly documented as only used if `specificProvider` is null:

>      * @param securityProvider
>      *            optional security provider to be used but only if there is no
>      *            specific provider defined for the specified scheme.

I think I'd be more confortable having this be opt-in behavior.

fernomac: Cool, I'll plumb that through.

aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/internal/crypto/AesGcm.java
```diff
@@ -49,7 +49,7 @@ CipherLite createAuxillaryCipher(SecretKey cek, byte[] ivOrig,
             NoSuchProviderException, NoSuchPaddingException,
             InvalidAlgorithmParameterException {
         byte[] iv = AES_CTR.adjustIV(ivOrig, startingBytePos);
-        return AES_CTR.createCipherLite(cek, iv, cipherMode, securityProvider);
+        return AES_CTR.createCipherLite(cek, iv, cipherMode, securityProvider, false);
```
fernomac: AES_CTR doesn't have a preferred provider, so 'false' is fine here.

aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/internal/crypto/CipherLite.java
```diff
@@ -86,7 +86,7 @@ private CipherLite() {
      */
     CipherLite recreate() {
         return scheme.createCipherLite(secreteKey, cipher.getIV(),
-                this.cipherMode, cipher.getProvider());
+                this.cipherMode, cipher.getProvider(), true);
```
fernomac: `true` to stick with the decision we made when creating the original cipher.

aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/internal/crypto/ContentCryptoScheme.java
```diff
@@ -227,7 +255,7 @@ CipherLite createCipherLite(SecretKey cek, byte[] iv, int cipherMode)
             throws InvalidKeyException, NoSuchAlgorithmException,
             NoSuchProviderException, NoSuchPaddingException,
             InvalidAlgorithmParameterException {
-        return createCipherLite(cek, iv, cipherMode, null);
+        return createCipherLite(cek, iv, cipherMode, null, false);
```
fernomac: IntelliJ suggests this method is never called. Maybe worth removing it?

fernomac:
@dagnir I added an opt-in flag.

fernomac:
(Travis says "Unresolveable build extension: Plugin org.apache.felix:maven-bundle-plugin:2.5.3 or one of its dependencies could not be resolved: Failed to read artifact descriptor for org.apache.felix:maven-bundle-plugin:jar:2.5.3: Could not transfer artifact org.apache.felix:maven-bundle-plugin:pom:2.5.3 from/to central (https://repo.maven.apache.org/maven2): Received fatal alert: protocol_version")

aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/internal/crypto/ContentCryptoScheme.java
```diff
@@ -198,6 +198,34 @@ CipherLite createCipherLite(SecretKey cek, byte[] iv, int cipherMode,
...
+
+        // Otherwise, if this crypto scheme prefers a particular provider (AesGcm prefers
+        // the non-FIPS BouncyCastle provider), that takes precedence.
+        if (preferredProvider != null) {
+            return Cipher.getInstance(algorithm, preferredProvider);
+        }
+
+        // Otherwise, if the user has specified a provider, go with that.
+        if (provider != null) {
+            return Cipher.getInstance(algorithm, preferredProvider);
```
dagnir: Is this correct? Seems like it should be `provider`, not `preferredProvider`

fernomac: Whoops! Nice catch!

fernomac:
@dagnir ping?

aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/model/CryptoConfiguration.java
```diff
@@ -233,8 +265,8 @@ public CryptoConfiguration withIgnoreMissingInstructionFile(
      *             specified crypto mode.
      */
     private void check(CryptoMode cryptoMode) {
-        if (cryptoMode == CryptoMode.AuthenticatedEncryption
-                || cryptoMode == CryptoMode.StrictAuthenticatedEncryption) {
+        if (cryptoProvider == null && (cryptoMode == CryptoMode.AuthenticatedEncryption
```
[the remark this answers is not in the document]
fernomac: To be strictly backwards compatible, yes, good catch. Change incoming.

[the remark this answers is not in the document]
shorea: Thanks, looks great!

fernomac:
👍It's Friday night, go home! :)

shorea:
Thanks for the contribution @fernomac! Merging this and it will be available in the next release.

PR #1639, opened 2018-06-20
Fix generic types on Return type for s3 waiters

jan-xyz:


aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/waiters/AmazonS3Waiters.java
```diff
@@ -50,7 +50,7 @@ public AmazonS3Waiters(AmazonS3 client) {
      * entered the desired state or not, where polling criteria is bound by
      * either default polling strategy or custom polling strategy.
      */
-    public Waiter bucketNotExists() {
+    public Waiter<HeadBucketRequest> bucketNotExists() {
```
dagnir: Can we all add the generic parameters for the acceptors? (Like `HttpFailureStatusAcceptor<HeadBucketResult>`?)

jan-xyz:
I pushed the changes for the `Http*StatusAcceptors`, there remains another warning, I'm not sure how to fix that.

jan-xyz:
I get `unchecked generic array creation for varargs parameter` on the `withAcceptor` call. I was thinking to solve it like it is solved in other SDKs (e.g. the [aws-java-sdk-emr](https://github.com/aws/aws-sdk-java/tree/4734de6fb0f80fe5768a6587aad3b9d0eaec388f/aws-java-sdk-emr/src/main/java/com/amazonaws/services/elasticmapreduce/waiters)), with one class per waiter state, but the code in there is generated. I could build it like that but I'm not sure if it's the right approach and on my first google I couldn't find how the code generator works.

dagnir:
Hmm, I think this actually safe so that should fine; we don't depend on the the type of the temporary array.

jan-xyz:
Okay, is there anything else I can do?

shorea:
Merged, thanks for the contribution!

PR #1665, opened 2018-07-09
Verifies a Cognito JSON Web Token (JWT)

### LINKED ISSUE #1664: Cognito JSON Web Token (JWT) Verification
The [Verifying a JSON Web Token](https://docs.aws.amazon.com/cognito/latest/developerguide/amazon-cognito-user-pools-using-tokens-verifying-a-jwt.html) docs describe several steps for verifying id and access JWT's. The security of a developer's app depends on implementing these instructions correctly and without any vulnerabilities. It's easy to miss one of these steps. Missing a step negates most of the security that Cognito brings to an external application. There are a [handful](https://forums.aws.amazon.com/thread.jspa?messageID=787330&#787330) of instances online where someone incorrectly implements java JWT verification. It's a little unrealistic to expect every developer to implement this verification correctly.

Google has officially supported JWT verification for Google Sign-In within their ["Using a Google API Client Library"](https://developers.google.com/identity/sign-in/web/backend-auth).

Please provide an officially supported implementation of JWT verification!

EthanLozano:
fixes #1664
Implements the instructions here:
https://docs.aws.amazon.com/cognito/latest/developerguide/amazon-cognito-user-pools-using-tokens-verifying-a-jwt.html

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ jdk:
   - oraclejdk8
 env:
   - MODULES_TO_SKIP=\!aws-java-sdk-osgi MAVEN_OPTS=-XX:MaxPermSize=256m
-  - MODULES_TO_SKIP=\!aws-java-sdk-osgi,\!aws-java-sdk-code-generator,\!aws-java-sdk-codegen-maven-plugin,\!aws-java-sdk-opensdk MAVEN_OPTS=-XX:MaxPermSize=256m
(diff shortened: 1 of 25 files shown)


shorea:
@EthanLozano this is really great but unfortunately we don't have the bandwidth to support another higher level library at this time. We may consider adding this to V2 if there is enough demand but for now I'd recommend releasing it on your own to Maven.

louis-jaris:
I'm interested in this feature, is there a discussion for that in SDK V2?
@EthanLozano I totally agree with you, this sort of verification should be handled in the SDK as long it's part of the AWS ecosystem.

PR #1673, opened 2018-07-13
Making SecureRandom configurable via CryptoConfiguration

fernomac:
As a follow-up to #1633, also making the `SecureRandom` instance used to generate CEKs and IVs configurable, so a FIPS-certified instance can be provided. This allows the use of the BC-FIPS provider in approved-only mode (at least for [the testing I've done](https://github.com/fernomac/s3-fips-encryption-demo)).


fernomac:
<3 thanks!

(p.s. any of you folks going to oscon this year?)

shorea:
@fernomac not that I know of but https://giphy.com/explore/nobody-tells-me-nothin

shorea:
Merged, thanks again for the contribution! :)

PR #1641, opened 2018-06-20
add getClientConfiguration accessor to AmazonWebServiceClient

dbyron0:
I'm looking for a way (without reflection) to see what configuration I actually ended up with, so I can log it.

PR #1677, opened 2018-07-16
change AmazonSNSClientBuilder and AmazonSQSClientBuilder return concr…

dbyron0:
…ete classes from build and defaultClient methods, instead of the interface

so callers have access to getClientConfiguration method.

Turns out https://github.com/aws/aws-sdk-java/pull/1641 wasn't enough.  It helps me get the configuration from the result of AmazonSNSClientBuilder.build() without reflection, but I still need a cast from AmazonSNS to AmazonSNSClient before the getClientConfiguration method is available.

This change makes that call possible with no cast.

### DIFF
--- a/aws-java-sdk-sns/src/main/java/com/amazonaws/services/sns/AmazonSNSClientBuilder.java
+++ b/aws-java-sdk-sns/src/main/java/com/amazonaws/services/sns/AmazonSNSClientBuilder.java
@@ -25,7 +25,7 @@ import com.amazonaws.client.AwsSyncClientParams;
  **/
 @NotThreadSafe
 @Generated("com.amazonaws:aws-java-sdk-code-generator")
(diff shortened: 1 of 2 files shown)


shorea:
These classes are generated and I'm not 100% sure if this is binary compatible. Ideally we'd add the new method to the interface instead of returning a concrete impl in the builder.

dbyron0:
OK, I can take a shot at adding the method to AmazonSNS and AmazonSQS, and implementing them as unsupported in AbstractAmazonSNS, AbstractAmazonSQS.  Seems like there's a kind of diamond pattern though e.g. AmazonSNSClient extends from AmazonWebServiceClient and implements AmazonSNS, and both of those will then have getClientConfiguration methods.

dbyron0:
How's this?  https://github.com/aws/aws-sdk-java/pull/1678

Still looks like AmazonSNS and AmazonSQS are generated files, but I couldn't find the templates for them.

shorea:
Closing in favor of #1678 . Will continue discussion there.

PR #1678, opened 2018-07-16
add getClientConfiguration method to AmazonSNS, AmazonSQS

dbyron0:
Turns out #1641 wasn't enough. It helps me get the configuration from the result of AmazonSNSClientBuilder.build() without reflection, but I still need a cast from AmazonSNS to AmazonSNSClient before the getClientConfiguration method is available.

This change makes that call possible directly on AmazonSNS.


shorea:
@dbyron0 just a quick update, we are discussing this change amongst the team and will let you know soon.

dbyron0:
Sounds good.  Thanks for the update.

shorea:
We have decided to incorporate this into the SDK but it may take several weeks to deal with some issues with internal customers implementing the interface. I'll close this PR for now as we need to make the change in the code generator and open an issue to track.

dbyron0:
Excellent.  Thanks again.

dbyron0:
Would love an update.  Any progress here?

PR #1705, opened 2018-07-31
Metadata ua update

willbengtson:
Update where the User-Agent gets set to catch the scenario missed as commented in  #1562


willbengtson:
Closing in favor of #1708

PR #1717, opened 2018-08-09
Throwing SdkClientException for the AWSStaticCredentialsProvider when…

fmcejudo:
… access key or secret key are null.

This was causing that when in AWSCredentialsProviderChain this provider was in place and the credential where empty accept them as valid.


spfink:
I'll investigate a bit further but I don't think it's possible for us to accept this PR as it could break backwards compatibility for other customers.

fmcejudo:
so @spfink, I was wondering if it was possible to have empty access key and secret key, by my understanding they are not, but I don't know if there is any use case where is contemplated. 

If this is not the fix, them the fix might be in `AWSCredentialsProviderChain` when it implements the `getCredential` as it lets those empty strings. What do you think?

PR #1722, opened 2018-08-14
Fix HTTP/HTTPS docstrings

tipabu:
They got swapped.


shorea:
Thanks for the fix, merged!

Issue #1704, opened 2018-07-31
Remove ThreadLocal use


davidmoten:
Following on from #934.

When the aws-sdk-java (any component) is used in a container like Glassfish or Tomcat a SEVERE/ERROR entry is written to the log whenever a webapp is bounced/redeployed or the container shutdown. Here's the error (glassfish):

```java
[2016-12-06T12:07:12.369-0500] [glassfish 4.1] [SEVERE] [] [javax.enterprise.web.util] [tid: _ThreadID=113 _ThreadName=admin-listener(6)] [timeMillis: 1481044032369] [levelValue: 1000] [[
  The web application [/XYZ] created a ThreadLocal with key of type [com.amazonaws.auth.AbstractAWSSigner$1] (value [com.amazonaws.auth.AbstractAWSSigner$1@39054dd0]) and a value of type [java.security.MessageDigest.Delegate] (value [SHA-256 Message Digest from SUN, <initialized>
]) but failed to remove it when the web application was stopped. Threads are going to be renewed over time to try and avoid a probable memory leak.]]

[2016-12-06T12:07:12.371-0500] [glassfish 4.1] [SEVERE] [] [javax.enterprise.web.util] [tid: _ThreadID=113 _ThreadName=admin-listener(6)] [timeMillis: 1481044032371] [levelValue: 1000] [[
  The web application [/XYZ] created a ThreadLocal with key of type [com.amazonaws.auth.SigningAlgorithm$1] (value [com.amazonaws.auth.SigningAlgorithm$1@396e15e3]) and a value of type [javax.crypto.Mac] (value [javax.crypto.Mac@411308ff]) but failed to remove it when the web application was stopped. Threads are going to be renewed over time to try and avoid a probable memory leak.]]
```
The ThreadLocals above are not actually memory leaks according to discussion in #918 and I am happy with the reasoning. 

However, we monitor for the existence of SEVERE log lines and raise on-call alerts when seen so I'd like to remove this red-herring so that I can monitor normally and so that we don't waste peoples time when they see the error report in their container logs.

[...]

PR #1729, opened 2018-08-21
Enable clearing of SDK ThreadLocals

davidmoten:
Discussed in #1704 

There are 5 uses of `ThreadLocal` in the SDK at the moment. 4 are in *aws-java-sdk-core* and 1 in *aws-java-sdk-sns*. Each instantiation has been wrapped with `SdkThreadLocalsRegistry.register()` (declared in an internal package).

When a user wants to clear `ThreadLocal` values specific to the SDK for the current thread he/she calls 

```java
SdkThreadLocals.remove();
```

cc @SuperEvenSteven @philipgroom

### DIFF
--- a/aws-java-sdk-core/src/main/java/com/amazonaws/internal/SdkThreadLocalsRegistry.java
+++ b/aws-java-sdk-core/src/main/java/com/amazonaws/internal/SdkThreadLocalsRegistry.java
@@ -1,0 +1,52 @@
+/*
+ * Copyright 2015-2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+package com.amazonaws.internal;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import com.amazonaws.annotation.SdkProtectedApi;
+
+@SdkProtectedApi
+public final class SdkThreadLocalsRegistry {
+    
+    private static final List<ThreadLocal<?>> threadLocals = new ArrayList<ThreadLocal<?>>();
+
+    private SdkThreadLocalsRegistry() {
+        // prevent instantiation
+    }
+    
+    /**
+     * Registers {@link ThreadLocal} objects in use by the AWS SDK so that their values can 
+     * be removed via the {@link #remove()} method.
+     * 
+     * <p>To avoid memory leaks and reduce contention this method should only be called when 
+     * setting static final locations (for example finals in enums or static final fields).
+     * 
+     * @param threadLocal ThreadLocal to register
+     * @return the input ThreadLocal
+     */
+    public static synchronized <T> ThreadLocal<T> register(ThreadLocal<T> threadLocal) {
+        threadLocals.add(threadLocal);
+        return threadLocal;
+    }
+
+    public synchronized static void remove() {
+        for (ThreadLocal<?> t: threadLocals) {
+            t.remove();
+        }        
+    }
+    
+}
--- a/aws-java-sdk-core/src/test/java/com/amazonaws/SdkThreadLocalsTest.java
+++ b/aws-java-sdk-core/src/test/java/com/amazonaws/SdkThreadLocalsTest.java
@@ -1,0 +1,51 @@
+/*
+ * Copyright 2015-2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+package com.amazonaws;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNull;
+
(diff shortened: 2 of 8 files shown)


davidmoten:
fixing case in classes (Sdk > SDK)

davidmoten:
Ah, I see there is no convention. Leaving as is.

davidmoten:
Unrelated CI failure:

```
[ERROR]   The project com.amazonaws:aws-java-sdk-osgi:1.11.392-SNAPSHOT (/home/travis/build/aws/aws-sdk-java/aws-java-sdk-osgi/pom.xml) has 1 error
[ERROR]     Unresolveable build extension: Plugin org.apache.felix:maven-bundle-plugin:2.5.3 or one of its dependencies could not be resolved: Failed to read artifact descriptor for org.apache.felix:maven-bundle-plugin:jar:2.5.3: Could not transfer artifact org.apache.felix:maven-bundle-plugin:pom:2.5.3 from/to central (https://repo.maven.apache.org/maven2): Received fatal alert: protocol_version -> [Help 2]
```

aws-java-sdk-core/src/test/java/com/amazonaws/SdkThreadLocalsTest.java
```diff
@@ -0,0 +1,37 @@
+package com.amazonaws;
```
zoewangg: missing copyright on this file

aws-java-sdk-core/src/main/java/com/amazonaws/internal/SdkThreadLocalsRegistry.java
```diff
@@ -0,0 +1,39 @@
...
+ * permissions and limitations under the License.
+ */
+package com.amazonaws.internal;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public final class SdkThreadLocalsRegistry {
+    
+    private static final List<ThreadLocal<?>> threadLocals = new ArrayList<ThreadLocal<?>>();
```
zoewangg: might be nice to use Map(className to ThreadLocal maybe?) here so that customers can remove ThreadLocal use for certain class only.

aws-java-sdk-core/src/test/java/com/amazonaws/SdkThreadLocalsTest.java
davidmoten: thanks, I'll fix

aws-java-sdk-core/src/main/java/com/amazonaws/internal/SdkThreadLocalsRegistry.java
davidmoten: I only have a use case (application shutdown/restart with a container managed thread pool) for removal of all. Can you suggest a use case for that scenario?

davidmoten:
I've added the missing copyright notice to the test class.

aws-java-sdk-core/src/main/java/com/amazonaws/internal/SdkThreadLocalsRegistry.java
```diff
@@ -0,0 +1,39 @@
...
+
+    private SdkThreadLocalsRegistry() {
+        // prevent instantiation
+    }
+    
+    public static synchronized <T> ThreadLocal<T> register(ThreadLocal<T> threadLocal) {
+        threadLocals.add(threadLocal);
+        return threadLocal;
+    }
```
millems: Right now, if `register` is used from a non-static final context, we could leak memory. Or if it's used on a critical path, we could be creating a performance bottleneck with the `synchronized` methods.

An alternative (more complex) implementation could fix those issues, if we see them as a problem:

```Java
class SdkThreadLocalsRegistry {
    private static final ThreadLocal<Set<WeakReference<ThreadLocal<?>>>> REGISTRY =
            ThreadLocal.withInitial(HashSet::new);
    private static final ThreadLocal<ReferenceQueue<Object>> RELEASED_REFERENCE_QUEUE =
            ThreadLocal.withInitial(ReferenceQueue::new);

    public static void register(ThreadLocal<?> threadLocal) {
        Set<WeakReference<ThreadLocal<?>>> threadLocalRegistry = REGISTRY.get();
        clearExpiredEntries(threadLocalRegistry);
        threadLocalRegistry.add(new WeakReference<>(threadLocal, RELEASED_REFERENCE_QUEUE.get()));
    }

    public static void remove() {
        REGISTRY.get().stream()
                .map(WeakReference::get)
                .filter(l -> l != null)
                .forEach(ThreadLocal::remove);
        REGISTRY.remove();
        RELEASED_REFERENCE_QUEUE.remove();
    }

    private static void clearExpiredEntries(Set<WeakReference<ThreadLocal<?>>> threadLocalRegistry) {
        Reference<?> releasedReference;
        while ((releasedReference = RELEASED_REFERENCE_QUEUE.get().poll()) != null) {
            threadLocalRegistry.remove(releasedReference);
        }
    }
}
```

This might introduce new issues I'm not aware of...

davidmoten: At the moment `register` will be called a maximum of 5 times because the ThreadLocal instances are class level singletons and only on creation of those singletons is `register` called. 

I think contention on the `SdkThreadLocals.remove()` method from multiple threads is unlikely because it would normally be called after a batch of single threaded IO interaction with AWS which will naturally limit the maximum call rate from a single thread. Perhaps I'm not being imaginative enough with expected usage?

Bear in mind that any addition of a ThreadLocal to the code base will have to trigger reviewers to suggest it be registered and at that point contention issues can be revisited.  I think the suggested replacement code adds unnecessary complexity for the moment given that a concrete use case (ThreadLocal instantiation that is not a class level singleton) has not been proposed for the sdk. The code you are suggesting may never be needed.

Having said all that, that's a nifty pattern you've suggested and I'm happy to review and include if desired.

davidmoten: > It's probably good to add a comment that register should only be used from static final locations (eg. finals in enums and static final variables).

Will do.

> Is this method used cross-module? If so, it would be good to mark it as an @SdkProtectedApi .

It's used in *aws-java-sdk-core* and *aws-java-sdk-sns* which I assume means it's cross module.  I'll add that annotation.

millems:
Thanks for your contribution! Can you confirm you're fine with all of your changes being released under the Apache 2.0 license?

davidmoten:
>Thanks for your contribution! Can you confirm you're fine with all of your changes being released under the Apache 2.0 license?

Yep, thanks for the teams efforts discussing and reviewing too.

PR #1760, opened 2018-09-12
InstanceProfileCredentialsProvider: use daemon thread for refresh.

vprus:
When refreshCredentialsAsync is specified, a scheduled executor
is used to refresh credentials, however that executor is not configured
to use daemon threads. As result, that non-daemon thread blocks JVM
exit. This change explicit make the thread into daemon.


dagnir:
LGTM! Can you confirm you're submitting this under the terms of the [Apache 2.0 License](https://github.com/aws/aws-sdk-java/blob/a96c560e4c4a290639a245784db85fd37b1117fd/LICENSE.txt#L1)?

vprus:
Yes, I am submitting this under Apache 2.0 license.

dagnir:
Thanks for the contribution! This will be available in the next release.

PR #1820, opened 2018-11-28
Early return in endpointDiscoveryEnabled

s12v:
Even when `endpointDiscoveryDisabled` is set to `true`, `DEFAULT_ENDPOINT_DISCOVERY_PROVIDER` still goes through default endpoint discovery providers, loads profiles, etc.

This PR changes `endpointDiscoveryEnabled` behavior to exit early.
In case when endpoint discovery is explicitly enabled or disabled, `endpointDiscoveryChainSetting` would be ignored anyways.

I couldn't find any tests for this functionality.


millems:
Sorry, I think we've actually fixed this one independent of this PR.

millems:
Actually, that was just V2 that I was looking at. It may not be fixed in V1.

s12v:
would it still make sense to fix it in v1 too, what do you think, @millems ?

debora-ito:
Closing this PR, we don't have plans to merge this at this moment.

Issue #1836, opened 2018-12-12
Support for region EU (Stockholm) / eu-north-1 not yet fully reflected in SDK


sopel:
While `eu-north-1` seems already properly reflected in [endpoints.json](https://github.com/aws/aws-sdk-java/blob/1.11.467/aws-java-sdk-core/src/main/resources/com/amazonaws/partitions/endpoints.json), it is still absent from the `Regions` enumeration in [Regions.java](https://github.com/aws/aws-sdk-java/blob/1.11.467/aws-java-sdk-core/src/main/java/com/amazonaws/regions/Regions.java#L33). This affects UX in terms of automatic display name resolution, and other than with #1417, it isn't easily worked around by using  [PartitionsLoader](https://github.com/aws/aws-sdk-java/blob/39ebf439b8e4050684cbfca1811c84b3ac5f2468/aws-java-sdk-core/src/main/java/com/amazonaws/partitions/PartitionsLoader.java#L45) capability to load an override `endpoints.json` from the class path .

* :question: While at it I might as well ask: Given the SDK is using a lot of code generation already, is there any reason why the enumeration is seemingly manually curated rather than automatically generated from the meanwhile available `partitions` list within [endpoints.json](https://github.com/aws/aws-sdk-java/blob/master/aws-java-sdk-core/src/main/resources/com/amazonaws/partitions/endpoints.json#L2), which would avoid such inconsistencies and delays?

Thanks!

spfink:
We've added support for this internally and it should be available after today's release. 

Do note that you should use the RegionUtils.getRegion(String region) method instead of the Regions or Region class directly:

https://github.com/aws/aws-sdk-java/blob/master/aws-java-sdk-core/src/main/java/com/amazonaws/regions/RegionUtils.java#L104

It is forwards compatible with all new regions even if we don't have the enum or endpoints.json updated.

spfink:
We do generate these enums in the v2 SDK:

https://github.com/aws/aws-sdk-java-v2

We could look into doing so for the v1 SDK as well but we would have to balance that effort against ongoing efforts to continue to build up the v2 codebase.

debora-ito:
The `eu-north-1` region is available in the latest release, so I'm going to go ahead and close this. Feel free to reopen if you have further questions.

PR #1839, opened 2018-12-13
Add Stockholm region

gardleopard:
`Caused by: java.lang.IllegalArgumentException: Cannot create enum from eu-north-1 value!`


gardleopard:
this relates to https://github.com/aws/aws-sdk-java/issues/1836

spfink:
This is already flowing through our release process and should be available after today's release.

Do note that you should use the RegionUtils.getRegion(String region) method instead of the Regions or Region class directly:

https://github.com/aws/aws-sdk-java/blob/master/aws-java-sdk-core/src/main/java/com/amazonaws/regions/RegionUtils.java#L104

It is forwards compatible with all new regions even if we don't have the enum or endpoints.json updated.

PR #1845, opened 2018-12-19
Deprecate AWS_REGION

tb3088:
No wonder there is so much confusion - apparently nobody has cross-checked the various APIs or against  the documentation. AWS_REGION hasn't been valid (per public docs) for a very long time.

### DIFF
--- a/aws-java-sdk-core/src/main/java/com/amazonaws/SDKGlobalConfiguration.java
+++ b/aws-java-sdk-core/src/main/java/com/amazonaws/SDKGlobalConfiguration.java
@@ -256,7 +256,7 @@ public class SDKGlobalConfiguration {
 
     /**
      * Environment variable to set an alternate path to the shared config file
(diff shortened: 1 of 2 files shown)


debora-ito:
Can you point out to the documentation that says AWS_REGION is not valid?

This [setup credentials doc](https://docs.aws.amazon.com/sdk-for-java/v1/developer-guide/setup-credentials.html#setup-credentials-setting-region) and this [region provider chain doc](https://docs.aws.amazon.com/sdk-for-java/v1/developer-guide/java-dg-region-selection.html#automatically-determine-the-aws-region-from-the-environment) show AWS_REGION as the environment variable name.

tb3088:
Amazon's documentation is riddled with old and deprecated information. The point is to update this (the oldest API implementation, and way back when AWS_REGION was indeed the variable) to track the revisions that have been incorporated into the API.

PR #1850, opened 2018-12-19
Removing unnecessary/unused imports

sergioescala:
Signed-off by: Sergio Escalante sergioescala@gmail.com


dagnir:
Thanks for the contribution! This has been pulled in upstream and should be included in the next release.

PR #1854, opened 2018-12-24
Add custom header to download transfer request, default value for uploads parameter, content-length for x-amzn-Remapped-Content-Length

olgunyldz:
1 - Custom header values should be copied to download request for large files. 
2 - while downloading large files via Amazon Api Gateway, uploads parameter will be needed for CreateMultipartUpload. but because of the null value, it cannot be parsed. I added a random value for this field.
3 - Api gateway overrite the content-length field, and add "x-amzn-Remapped-Content-Length" for the file size.


millems:
Sorry that this went so long without a reply. We're nervous about making non-trivial changes for 1.11.x at this time, because we've found that even seemingly-safe changes can affect a handful of users and pull us away from development on 2.x of the SDK. Because of this, we will not be merging this pull request in 1.11.x.

Issue #1863, opened 2019-01-04
Base64 static initialization fails when jaxb-core not in classpath with jaxb-impl >= 2.2.11


akhaku:
#1428 noted that certain classes were moved from jaxb-impl to jaxb-core, which caused Base64 to fail to statically initialize. ba4a6a4ce49 updated the static initializer to attempt to not fail if jaxb was not enabled in Java 9. However, in Java8, if you have jaxb-impl on your classpath but not jaxb-core, `JAXBContext.newInstance()` fails with a `NoClassDefFoundError` (which wraps a `ClassNotFoundException`), meaning that the exception is not caught and Base64 fails to initialize.

<img width="1040" alt="screen shot 2019-01-04 at 2 56 21 pm" src="https://user-images.githubusercontent.com/463176/50716773-0b946b80-1038-11e9-941a-1c9668b2122a.png">

varunnvs92:
Isn't `jaxb-core` required to properly initialize JAXBContext?
Can you clarify whats the expected behavior should be?

akhaku:
As of jaxb-impl >= 2.2.11, yes. It looks like the intention of the static block in Base64 is to use jaxb if possible, but fall back to the SDK impl if it's not available. Not catching the NoClassDefFoundError makes that second bit break - the expected behaviour in that case is to bail out and use the SDK impl. With jaxb-impl < 2.2.11, we don't get the NoClassDefFoundError since only jaxb-impl is needed and not jaxb-core. You're right that we should be including the jaxb-core jar ourselves, but Base64 probably shouldn't fail to initialize without that jar.

varunnvs92:
Got it. Catching `Throwable` instead of `Exception` should fix the issue. I will make the change.

akhaku:
Thanks! I opened #1864 to make that change - might be simpler to hit merge on that guy?

varunnvs92:
Fixed in v1.11.485. Thanks for reporting the issue and PR.

brunoeufrasio:
> Corrigido em v1.11.485. Obrigado por relatar o problema e PR.

Atualizando pra essa versão resolveu pra mim tambem

PR #1864, opened 2019-01-04
Catch Throwable to deal with NoClassDefFoundError (#1863)

akhaku:
Not having certain classes on the classpath results in a
NoClassDefFoundError (that wraps a ClassNotFoundException)
which causes static initialization failure. Protect against
that by catching Throwable instead of Exception.


varunnvs92:
Sorry for the delay. Can you confirm that you are submitting the code under Apache 2.0 license?

akhaku:
I can confirm that, yes. I am submitting the code under the Apache 2.0 license.

varunnvs92:
Made the change internally with only catching `NoClassDefFoundError` in addition to Exception. This is to be safe we don't suppress errors unrelated to class finding.
The change will be released in next release (by end of week).

Issue #1872, opened 2019-01-09
JsonPolicyReader does not support the NotResource element


mhuffman-r7:
`JsonPolicyReader` does not support the `NotResource` resource node in policy statements. However, as of 1.11.298 `JsonPolicyWriter` supported writing `NotResource`. Strangely enough the `withIsNotType` is unused outside of unit tests. The reader simply needs to be updated to support parsing and adding these resources to the statement object.

https://github.com/aws/aws-sdk-java/blob/master/aws-java-sdk-core/src/main/java/com/amazonaws/auth/policy/internal/JsonPolicyReader.java#L149

I would like this issue to be assigned to me.

mhuffman-r7:
Is this likely to be accepted any time soon?

debora-ito:
Feature released in `1.11.580`. Thank you.

mhuffman-r7:
Thank you.

PR #1873, opened 2019-01-09
Add support for NotResource to JsonPolicyReader

mhuffman-r7:
Adds `NotResource` parsing in statements to `JsonPolicyReader`. Unit test added.

Fixes https://github.com/aws/aws-sdk-java/issues/1872


aws-java-sdk-core/src/main/java/com/amazonaws/auth/policy/internal/JsonPolicyReader.java
```diff
@@ -146,9 +146,17 @@ private Statement statementOf(JsonNode jStatement) {
...
 
+        List<Resource> resources = new LinkedList<Resource>();
         JsonNode resourceNodes = jStatement.get(JsonDocumentFields.RESOURCE);
         if (isNotNull(resourceNodes))
-            statement.setResources(resourcesOf(resourceNodes));
+            resources.addAll(resourcesOf(resourceNodes, false));
+
+        JsonNode notResourceNodes = jStatement.get(JsonDocumentFields.NOT_RESOURCE);
+        if (isNotNull(notResourceNodes))
+            resources.addAll(resourcesOf(notResourceNodes, true));
```
varunnvs92: Is it valid to have both `RESOURCE` and `NOT_RESOURCE` type in same policy? If so, can you add more tests having both types?

spfink:
This went out in 1.11.580.

Thanks for the contribution!

PR #1899, opened 2019-01-28
Synchronize overridden synchronized methods

seanmcl:
All the below overridden methods are marked `synchronized` in the
OpenJDK.  If a synchronized method is overridden in a subclass, and
the overriding method is not synchronized, the thread-safety of the
subclass may be broken.

Since there are no thread-safety comments in these files, I'm not sure
you'll want to merge these.  If we decide not to merge though, it
would be great to document that the classes are no longer thread safe.

### DIFF
--- a/aws-java-sdk-core/src/main/java/com/amazonaws/event/ProgressInputStream.java
+++ b/aws-java-sdk-core/src/main/java/com/amazonaws/event/ProgressInputStream.java
@@ -164,7 +164,7 @@ public abstract class ProgressInputStream extends SdkFilterInputStream {
     }
 
     @Override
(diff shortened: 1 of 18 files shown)


millems:
Looking at these changes, it looks like it's mostly documentation changes that we need to make (specifically, the input streams aren't thread safe). The ones that delegate are thread safe, since the delegate will be synchronized. 

While I agree we should probably update the documentation, we're focusing most of our development effort on V2 of the Java SDK and seen people being bitten bitten by these not being documented as not-thread-safe. 

My inclination is just to close this issue and make any changes in V2 if customers express confusion. Do you disagree?

PR #1904, opened 2019-02-04
Replace 'joda-time' with Java 8

stepio:
Hello,

Please don't kick me.

Let's consider this a POC to address https://github.com/aws/aws-lambda-java-libs/issues/72

In short:
- updated JDK version to 8;
- removed joda-time;
- replaces DateTime with ZonedDateTime;
- fixed everything that got broken.


millems:
Thanks for all the effort you put into this, but unfortunately Java 6 is a hard requirement of this project. We've migrated to Java 8 and replaced joda-time in 2.x of the SDK: https://github.com/aws/aws-sdk-java-v2. We won't be able to merge this PR.

If you're having any issues with joda-time, maybe you could consider migrating to 2.x of the SDK? It doesn't support every feature of 1.11.x at this time, unfortunately.

stepio:
Thanks @millems , I didn't know that you already have an ongoing work for Java 8 migration.

But why is https://github.com/aws/aws-sdk-java-v2 not integrated to https://github.com/aws/aws-lambda-java-libs then?

Does not make sense to me...

millems:
Unfortunately, that's under the "it doesn't support every feature of 1.11.x at this time" umbrella. 2.x isn't backward-compatible with 1.11.x, so the migration path isn't as straight-forward. 

It'll take a while for the ecosystem to catch up to 2.x, since 1.x was out for 9 years, and 2.x has been out for 2 months.

PR #1924, opened 2019-02-26
Incorrect javadoc on networkInterfaces field.

ttwd80:
Documentation on RunInstancesRequest.networkInterfaces is incorrect

The documentation that says `You cannot specify this option and the network interfaces option in the same request.` should not be associated with `networkInterfaces`. You want to specify a network interface, but the documentation is saying if you use this option, you can't use this option.

I think the documentation is accurate for  `ipv6AddressCount`, `ipv6Addresses`, `securityGroupIds`, `securityGroups`, `subnetId` and `privateIpAddress`. But not for `networkInterfaces`.


debora-ito:
Hi @ttwd80, thank you for the PR. The javadocs are code generated from service models provided by the service team, and the issues you pointed out were fixed recently, so I'll go ahed and close this.

Issue #1862, opened 2019-01-03
DAX Request Handlers


nirajpatel:
Hi! Using the `AmazonDynamoDBClientBuilder` we have the ability to specify request handlers, but the `AmazonDaxClientBuilder` does not have that ability as the class does not extend `AwsClientBuilder`. Is being able to set up request handlers on the road map? We lose the ability to read capacity usage from DAX/DDB requests.

debora-ito:
Hi @nirajpatel, assuming you're talking about the `AmazonDaxClientBuilder` from the [DAX Client](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/DAX.client.html), I'll forward your request to the DAX team, the AWS SDK team doesn't own the DAX Client. 

Going to close this issue but will post any updates we get from the DAX team.

nirajpatel:
Yes, that client is correct. Thank you @debora-ito!

nirajpatel:
> You can set the ReturnConsumedCapacity property on the request objects such as GetItemRequest to return the read capacity usage. I suppose their usage is not to do it at the request level, but at the client level using something like https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/monitoring/internal/ClientSideMonitoringRequestHandler.html.

Okay, we will look to see if we can hook into the usage at the response level. We are using the DynamoDBMapper so I think we might have to add a layer on top of that. Thank you for getting back to me on this issue! I will reach back out if we have any questions.

PR #1934, opened 2019-03-11
Add ability to specify default return consumed capacity in DynamoDBMapper

nirajpatel:
**Issue:** https://github.com/aws/aws-sdk-java/issues/1862

**Description of changes:** Adds the ability to specify a default `ReturnConsumedCapacity` value for operations using the `DynamoDBMapper`. We don't have the ability to set the `ReturnConsumedCapacity` on individual DynamoDB operations as they are built up inside of the DynamoDBMapper. We currently get around this when using DynamoDB by using reflection and setting a `ReturnConsumedCapacity` on all requests; however, there are no pre-request hooks available when using DAX. This will allow us to move forward with DAX without losing the ability to track consumed capacity.


millems:
Unfortunately we're concentrating our development effort on 2.x of the SDK, and as such won't be adding more features to the 1.11.x mapper. Sorry about that.

PR #1952, opened 2019-03-27
Get sublist of PaginatedList w/o loading full set

patrickdeshpande:
We should not have to load the entire result set in order to determine the upper bound for the sublist method. This is currently necessary because we do not know if our upper bound index is out of range of the result set.

*Description of changes:*


aws-java-sdk-dynamodb/src/main/java/com/amazonaws/services/dynamodbv2/datamodeling/PaginatedList.java
```diff
@@ -394,6 +394,24 @@ public boolean contains(Object arg0) {
         return Collections.unmodifiableList(allResults.subList(arg0, arg1));
     }
 
+    /**
+     * Returns a sub-list in the range specified, loading more results as
+     * necessary. May return a list of size less than (arg1 - arg2)
```
patrickdeshpande: arg2 -> arg0*

patrickdeshpande:
No point in keeping this open

Issue #1816, opened 2018-11-26
AmazonKinesisFirehoseAsyncClient: Unable to unmarshall exception content


pushkar-g:
We are sporadically (once a day) NullPointerException the below error in production while calling AWS Firehose Client. 

**SDK Version -** 1.11.455
**JVM Version -** 8u181-jre
**AWS Region-** us-west-2

> java.lang.NullPointerException: null
at com.fasterxml.jackson.databind.ObjectMapper.treeToValue(ObjectMapper.java:2745) ~[jackson-databind-2.9.7.jar:2.9.7]
at com.amazonaws.transform.JsonErrorUnmarshaller.unmarshall(JsonErrorUnmarshaller.java:52) ~[aws-java-sdk-core-1.11.455.jar:na]
at com.amazonaws.http.JsonErrorResponseHandler.unmarshallException(JsonErrorResponseHandler.java:108) [aws-java-sdk-core-1.11.455.jar:na]
at com.amazonaws.http.JsonErrorResponseHandler.createException(JsonErrorResponseHandler.java:96) [aws-java-sdk-core-1.11.455.jar:na]
at com.amazonaws.http.JsonErrorResponseHandler.handle(JsonErrorResponseHandler.java:63) [aws-java-sdk-core-1.11.455.jar:na]
at com.amazonaws.http.JsonErrorResponseHandler.handle(JsonErrorResponseHandler.java:33) [aws-java-sdk-core-1.11.455.jar:na]
at com.amazonaws.http.AwsErrorResponseHandler.handleAse(AwsErrorResponseHandler.java:53) [aws-java-sdk-core-1.11.455.jar:na]
at com.amazonaws.http.AwsErrorResponseHandler.handle(AwsErrorResponseHandler.java:41) [aws-java-sdk-core-1.11.455.jar:na]
at com.amazonaws.http.AwsErrorResponseHandler.handle(AwsErrorResponseHandler.java:26) [aws-java-sdk-core-1.11.455.jar:na]
at com.amazonaws.http.AmazonHttpClient$RequestExecutor.handleErrorResponse(AmazonHttpClient.java:1677) [aws-java-sdk-core-1.11.455.jar:na]
at com.amazonaws.http.AmazonHttpClient$RequestExecutor.executeOneRequest(AmazonHttpClient.java:1350) [aws-java-sdk-core-1.11.455.jar:na]
[...]

PR #1954, opened 2019-03-28
Firehose Unable to unmarshall exception content

alphafoobar:
- Discovered 503 errors did not have json content which caused a
NullPointerException to be thrown and logged consuming our support and
development time. Instead of throwing an exception, we are checking for
null or NullNode before attempting to deserialize the json node to an
exception.
- Additionally the Amazon request header that was being sought was not the
same as the header that is contained in these errors.

See issue: https://github.com/aws/aws-sdk-java/issues/1816

*Issue #1816 *

*Description of changes:*

1. Added null check to JsonErrorResponseHandler
1. Added support for alternative Amazon request Id header as seen in 503 error responses
1. Added additional unit testing around these scenarios


millems:
Thanks for your contribution. I wanted to apologize for how long it's taking us to review it.

zoewangg (approved):
Thank you for your contribution! We have imported the change internally.

PR #1955, opened 2019-03-28
Support the Apache http client  connection manager shared mode setting

Crab2died:
Support the Apache http client connection manager shared mode setting.


millems:
Hi, unfortunately we are focusing most of our development effort on 2.x of the SDK and as such won't be adding many new features to 1.11.x of the SDK.

Issue #1919, opened 2019-02-20
org.apache.httpcomponents:httpclient:4.5.7 breaks fetching S3 objects with consecutive slashes in the key


akhaku:
org.apache.httpcomponents:httpclient:4.5.7 made a change[1] to url rewriting (used when executing a request) where it normalizes the provided urls by removing empty path segments (effectively collapsing consecutive slashes into a single slash).

S3 supports objects with consecutive slashes, but because of this behaviour change in httpclient, you can no longer fetch those objects with the AWS SDK, since the url httpclient makes a request to is different. On a side note, the _manifestation_ of this issue is a signature mismatch, since the signature calculated by the SDK uses double slashes but when httpclient hits the path without double slashes, the calculated signature is different.

It looks like there's some debate on it now and they're actually working on making normalization optional[2], so once that's released we can update the RequestConfig[3] to disable normalization and revert to the previous behaviour[3].

Example stack trace showing the code path to slash collapsing:
```
rewriteURI:158, URIUtils (org.apache.http.client.utils)
rewriteURIForRoute:231, URIUtils (org.apache.http.client.utils)
rewriteRequestURI:95, ProtocolExec (org.apache.http.impl.execchain)
execute:132, ProtocolExec (org.apache.http.impl.execchain)
doExecute:185, InternalHttpClient (org.apache.http.impl.client)
execute:83, CloseableHttpClient (org.apache.http.impl.client)
execute:56, CloseableHttpClient (org.apache.http.impl.client)
execute:72, SdkHttpClient (com.amazonaws.http.apache.client.impl)
executeOneRequest:1285, AmazonHttpClient$RequestExecutor (com.amazonaws.http)
executeHelper:1101, AmazonHttpClient$RequestExecutor (com.amazonaws.http)
doExecute:758, AmazonHttpClient$RequestExecutor (com.amazonaws.http)
executeWithTimer:732, AmazonHttpClient$RequestExecutor (com.amazonaws.http)
execute:714, AmazonHttpClient$RequestExecutor (com.amazonaws.http)
access$500:674, AmazonHttpClient$RequestExecutor (com.amazonaws.http)
execute:656, AmazonHttpClient$RequestExecutionBuilderImpl (com.amazonaws.http)
execute:520, AmazonHttpClient (com.amazonaws.http)
invoke:4705, AmazonS3Client (com.amazonaws.services.s3)
invoke:4652, AmazonS3Client (com.amazonaws.services.s3)
[...]

PR #1966, opened 2019-04-07
Don't allow httpclient to normalize URIs #1919

akhaku:
#1919

Set option in HttpClient RequestConfig to not normalize URIs. This was the default behavior before httpclient 4.5.7.

### DIFF
--- a/aws-java-sdk-core/src/main/java/com/amazonaws/http/apache/utils/ApacheUtils.java
+++ b/aws-java-sdk-core/src/main/java/com/amazonaws/http/apache/utils/ApacheUtils.java
@@ -28,6 +28,7 @@ import org.apache.http.auth.Credentials;
 import org.apache.http.auth.NTCredentials;
 import org.apache.http.client.AuthCache;
 import org.apache.http.client.CredentialsProvider;
+import org.apache.http.client.config.RequestConfig;
 import org.apache.http.client.methods.HttpRequestBase;
 import org.apache.http.client.protocol.HttpClientContext;
 import org.apache.http.entity.BufferedHttpEntity;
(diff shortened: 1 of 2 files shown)


varunnvs92:
Thanks for the PR. We are aware of the Apache behavior change. We opened a [ticket](https://issues.apache.org/jira/browse/HTTPCLIENT-1983) with Apache to change disable normalization by default. If Apache agrees to make the change, we don't need any change in the SDK and customers should be able to upgrade to newer Apache versions. Once a decision is taken in the ticket, We will make the appropriate changes in the SDK. Thank you for your patience.

akhaku:
Thanks! It doesn't look like that ticket is really moving along :( I share your concern about breaking people at runtime, tweaked the PR to not fail at runtime for people with older httpclients.

varunnvs92:
Thank you for updating the PR. The Apache ticket is closed. We will make changes in the SDK. Due to how the SDK code is sourced, I need to make this change internally before I can merge the PR or it will get overridden at next release. This is top priority item for us and will make a release this week with the necessary changes.

akhaku:
Works for me - feel free to merge or decline as needed! Thank you very much.

varunnvs92:
Found a bug with Apache with 4.5.8 which is a blocker to make this change.
https://issues.apache.org/jira/browse/HTTPCLIENT-1986

Stephan202:
[HTTPCLIENT-1986](https://issues.apache.org/jira/browse/HTTPCLIENT-1986) was fixed in Apache HTTP Client 4.5.9, released June 7. IIUC that unblocks this PR :tada:.

varunnvs92:
@Stephan202 Thanks for notifying us. Based on backlog prioritization, we have scheduled the work for next week.

akhaku:
Rebased and bumped the required httpclient version to 4.5.9, which has the pathing fix Varun talked about.

michael-o:
Committer here. This is not a bug in HttpClient, but solely in the S3 server stack which AWS must fix. You are just fixing the symptom. Read the upstream ticket.

debora-ito:
Hey @akhaku thank you for the contribution! 

In version `1.11.596` we released a solution for the `httpclient` problem, so I'll go ahead and close this PR.

Issue #1959, opened 2019-04-02
Use environment variables proxy settings 


benkropf:
Following up to #596, feel free to close this if this has already been implemented, but from what I can tell it has not.

ClientConfiguration should check common environment variables `HTTP_PROXY`, `HTTPS_PROXY` and `NO_PROXY` if neither configured ClientConfiguration variables or Java system properties are set. Precedence would be configured values, then Java system properties, then environment variables.

I'd be happy to have this assigned to me as well if this is something we'd like implemented.

webbhlin:
+1
I would like to see someone fixes this issue.

debora-ito:
Hi @benkropf, this seems a reasonable ask, marking as a feature request. If you can send us a PR, we are happy to take a look.

debora-ito:
Closed via #1967.

PR #1967, opened 2019-04-07
Use environment variables proxy settings

benkropf:
#1959 

*Description of changes:*
Check if common environment variables `HTTPS_PROXY`/`HTTP_PROXY`/`https_proxy`/`http_proxy` are set if neither the configured variables or Java properties are set. Precedence is configured values, then Java system properties, then environment variables.

Also checks for non proxy hosts using `NO_PROXY` and `no_proxy`, again keeping same precedence as above.

### DIFF
--- a/aws-java-sdk-core/src/main/java/com/amazonaws/ClientConfiguration.java
+++ b/aws-java-sdk-core/src/main/java/com/amazonaws/ClientConfiguration.java
@@ -1078,24 +1078,35 @@ public class ClientConfiguration {
     /**
      * Returns the Java system property for nonProxyHosts. We still honor this property even
      * {@link #getProtocol()} is https, see http://docs.oracle.com/javase/7/docs/api/java/net/doc-files/net-properties.html.
+     * This method expects the property to be set as pipe separated list.
      */
     private String getNonProxyHostsProperty() {
         return getSystemProperty("http.nonProxyHosts");
     }
 
     /**
-     * Returns the value of the environment variable NO_PROXY/no_proxy.
+     * Returns the value of the environment variable NO_PROXY/no_proxy. This method expects
+     * the environment variable to be set as a comma separated list, so this method
+     * converts the comma separated list to pipe separated list to be used internally.
      */
     private String getNonProxyHostsEnvironment() {
-        return getEnvironmentVariableCaseInsensitive("NO_PROXY");
+        String nonProxyHosts = getEnvironmentVariableCaseInsensitive("NO_PROXY");
+        if (nonProxyHosts != null) {
+            nonProxyHosts = nonProxyHosts.replace(",", "|");
+        }
+
+        return nonProxyHosts;
     }
 
     /**
      * Returns the optional hosts the client will access without going
      * through the proxy. Returns either the nonProxyHosts set on this
      * object, or if not provided, returns the value of the Java system property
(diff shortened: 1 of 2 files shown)


aws-java-sdk-core/src/main/java/com/amazonaws/ClientConfiguration.java
```diff
@@ -616,6 +618,48 @@ private String getSystemProperty(String property) {
...
+     * Returns the value for the given environment variable.
+     */
+    private String getEnvironmentVariable(String variable) {
+        return System.getenv(variable);
+    }
+
+    /**
+     * Returns a URL object for the given environment variable, if can be constructed, null otherwise.
+     */
+    private URL getUrlFromEnvironmentVariable(String variable) {
```
spfink: Can we rename variable to environmentVariable? Here and in all new places.

aws-java-sdk-core/src/main/java/com/amazonaws/ClientConfiguration.java
```diff
@@ -655,18 +699,49 @@ private String getProxyHostProperty() {
...
+        if (getProtocol() == Protocol.HTTPS) {
+            if (getHttpsProxyUrlFromEnvironment() != null) {
+                return getHttpsProxyUrlFromEnvironment().getHost();
+            }
+        } else {
+            if (getHttpProxyUrlFromEnvironment() != null) {
+                return getHttpProxyUrlFromEnvironment().getHost();
+            }
+        }
+        return null;
```
spfink: Can you change the logic here to match the system property logic? For example, not returning null here and catching it higher.

aws-java-sdk-core/src/main/java/com/amazonaws/ClientConfiguration.java
benkropf: Good catch, fixed in the newest commit

aws-java-sdk-core/src/main/java/com/amazonaws/ClientConfiguration.java
benkropf: Good point, I felt like this logic was a little clunky. The complication came from supporting both `HTTP_PROXY` and `http_proxy` so the logic was a little more complicated than the system property logic. The newest commit should clean it up a bit. Let me know what you think

benkropf:
@debora-ito can you take a look at this?

benkropf:
+@dagnir

dresnick-sf:
Could this please be merged to master and released? @spfink

sf-mkirilov:
I think that this PR should include some guidelines about the NO_PROXY format env variable. For example, curl defines it as comma-separated list of hosts while the SdkProxyRoutePlanner expects them to be '|' separated.

benkropf:
@sf-mkirilov good point, right now its expecting a `|` separated list because that was the standard defined by Oracle [here](http://docs.oracle.com/javase/7/docs/api/java/net/doc-files/net-properties.html). However, I think following the same convention as curl makes more sense for environment variables, since that's what GNU defines [here](http://www.gnu.org/software/wget/manual/html_node/Proxies.html) I'm open to suggestions on this one, but for the time being, I'll change it to be a comma separated list and document that.

spfink:
Merged and will go out with tomorrow's release.

PR #1984, opened 2019-04-29
Fix KMS docs typos

iamyohann:
*Description of changes:*
Fix KMS docs typos

There are some typo's over here too:
https://docs.aws.amazon.com/cli/latest/reference/kms/generate-data-key.html

Please let me know if any other repo's need updating?


dagnir:
Hi @iamyohann thank you for the PR. Unfortunately these javadocs are code generated from the [service models](https://github.com/aws/aws-sdk-java/blob/master/aws-java-sdk-models/src/main/resources/models/kms-2014-11-01-model.json) provided by the service team so it will need to be fixed there. We will pass the feedback along to the service team so they can fix it. I'm going to go ahead and close this.

juneb:
Thanks for the feedback. 

These typos all have the same source in the TrentDocumentationModel. I fixed the error on 7/27/19:
https://code.amazon.com/packages/TrentDocumentationModel/commits/f1ca5cddc6dc9cbbf6bb64e8950f8a34a5d8206b#

I'm not sure about Trebuchet tickets in the interim, but we had a Trebuchet update that used this commit on 11/08, which had the fix for the typo.
https://code.amazon.com/packages/TrentDocumentationModel/blobs/025d02aef253545d5f0bd34f0c335329bb822232/--/model/documentation.xml#L125

I've sent a query to the SDK writers for Java and the CLI to find out why the docs are not updated. I'll update this ticket with their reply.

PR #2030, opened 2019-06-24
Rethrowing an original IOException within SdkClientException

team3:
*When IOException occurs while reading file content, original exception must be populated within custom SdkClientException*


varunnvs92:
Thank you for the contribution. This will be included in the next release.

Issue #2080, opened 2019-08-19
S3EventNotificationRecord to have an S3Event enum as well as eventName


juanitosvq:
The class _S3EventNotificationRecord_ contains a property called _eventName_ of the type String. In my application, I need to have different logic depending on the type of event. I am going to have to hard code the different values this event can have, but I could use the enum in _S3Event_ for this. Something like the following:

```
public static class S3EventNotificationRecord {

        private final String awsRegion;
        private final String eventName;
        private final S3Event eventType;
```

Another solution would be to have a **fromString** method in the S3Event, so at least I could convert the incoming value in the record into the enum. However, the incoming values don't match exactly the values in the enum: _ObjectRestore:Completed_ vs _s3:ObjectRestore:Completed_ so the first solution would be preferred. This would be something like:

`
S3Event s3Event = S3Event.fromString("s3:ObjectRestore:Completed");
`

juanitosvq:
I am looking at adding this.

juanitosvq:
Any feedback on the PR open for this issue?

juanitosvq:
Closing after PR was merged

PR #2081, opened 2019-08-20
Add fromString to S3Event enum and include S3Event in S3EventNotificationRecord

juanitosvq:
#2080 

Include an S3Event object inside an S3EventNotification when parsing the JSON sent by an S3 event, instead of just a string with the event name. 
Also, add _fromString_ method to S3Event enum to make easier decoding the value. Given that S3 events return strings like _ObjectRestore:Completed_ but the values in the enum look like _s3:ObjectRestore:Completed_, I decided to include the prefix if no values match.

### DIFF
--- a/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/event/S3EventNotification.java
+++ b/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/event/S3EventNotification.java
@@ -330,6 +330,7 @@ public class S3EventNotification {
 
         private final String awsRegion;
         private final String eventName;
+        private final S3Event s3Event;
         private final String eventSource;
         private DateTime eventTime;
         private final String eventVersion;
@@ -343,6 +344,7 @@ public class S3EventNotification {
         public S3EventNotificationRecord(
                 String awsRegion,
                 String eventName,
+                S3Event s3Event,
                 String eventSource,
                 String eventTime,
                 String eventVersion,
@@ -353,6 +355,7 @@ public class S3EventNotification {
         {
             this(awsRegion,
                  eventName,
+                 s3Event,
                  eventSource,
                  eventTime,
                  eventVersion,
@@ -367,6 +370,7 @@ public class S3EventNotification {
         public S3EventNotificationRecord(
                 @JsonProperty(value = "awsRegion") String awsRegion,
                 @JsonProperty(value = "eventName") String eventName,
+                @JsonProperty(value = "s3Event") S3Event s3Event,
                 @JsonProperty(value = "eventSource") String eventSource,
                 @JsonProperty(value = "eventTime") String eventTime,
                 @JsonProperty(value = "eventVersion") String eventVersion,
@@ -379,6 +383,7 @@ public class S3EventNotification {
             this.awsRegion = awsRegion;
             this.eventName = eventName;
             this.eventSource = eventSource;
+            this.s3Event = S3Event.fromString(eventName);
 
             if (eventTime != null)
             {
@@ -399,6 +404,10 @@ public class S3EventNotification {
 
         public String getEventName() {
             return eventName;
+        }
+        
+        public S3Event getS3Event() {
+        	return s3Event;
         }
 
         public String getEventSource() {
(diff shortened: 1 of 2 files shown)


juanitosvq:
Any feedback on this?

aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/model/S3Event.java
```diff
@@ -53,4 +57,22 @@ private S3Event(String event) {
...
+    
+	private static final Map<String, S3Event> stringToEnum = new HashMap<>();
+
+	static {
+		for(S3Event s3Event : S3Event.values()) {
+			stringToEnum.put(s3Event.toString().toUpperCase(), s3Event);
+		}
+	}
+
+	public static S3Event fromString(String key) {
```
dagnir: Our other `enum` use `fromValue`, can we use the same name here?

Let's also follow the same basic logic used in the generated enums: https://github.com/aws/aws-sdk-java/blob/0549ecc7b625dbe1be44169a874408e241c8d1d8/aws-java-sdk-api-gateway/src/main/java/com/amazonaws/services/apigateway/model/SecurityPolicy.java#L37-L59

aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/event/S3EventNotification.java
```diff
@@ -343,6 +344,7 @@ public String getLifecycleRestoreStorageClass() {
         public S3EventNotificationRecord(
                 String awsRegion,
                 String eventName,
+                S3Event s3Event,
```
dagnir: Adding a new constructor parameter would be a breaking change unfortunately. The way around this is to just create a new constructor that takes `s3Event` as an additional parameter and deprecate the existing ones.

juanitosvq: Thanks for the feedback, @dagnir, this comment has been addressed in the new commit. That constructor was already deprecated, so I just added a new one.

aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/model/S3Event.java
juanitosvq: Thanks for the feedback, @dagnir. I changed this in the new commit, following the same logic suggested. I kept the extra bit where if the `value` doesn't match, I give it another chance by prefixing it with `s3:`. This was because the events receiving from S3 are like this:
`"eventName": "ObjectRestore:Completed"`

dagnir:
@juanitosvq Thanks for making the changes!

So after looking at this more closely, can you tell me why `S3Event` is being added to the `S3EventNotificationRecord`?  It seems a little redundant (and also lossy) since the `eventName` member exists, which has the exact member event name, like `s3:ObjectCreated:Put`.

juanitosvq:
@dagnir thanks for looking into this:
In our use case, we are parsing the incoming JSON from an S3 event into an `S3EventNotification` object. Our application needs to apply different logic based on the type of event, I thought it would make sense for the `S3EventNotification` to already include the enum instead of my application having to store `eventName` and then convert it into the `S3Event` enum. Ideally we would remove the property `String eventName` and instead only contain `S3Event s3Event`, but that would break backwards compatibility, that's the reason why the information will be duplicate now.

So instead of this:
```
S3EventNotification eventNotification = S3EventNotification.parseJson(message);
List<S3EventNotificationRecord> records = eventNotification.getRecords();

for (S3EventNotificationRecord record : records) {
	String eventName = record.getEventName();
	S3Event s3Event = S3Event.fromValue(eventName);
	switch(s3Event) {
		case ObjectRestoreCompleted:
			// do something
			break;
		case ObjectRestorePost:
			// do something else
	}
}
```

... I think it makes sense we can do this:
```      
S3EventNotification eventNotification = S3EventNotification.parseJson(message);
List<S3EventNotificationRecord> records = eventNotification.getRecords();

for (S3EventNotificationRecord record : records) {
	S3Event s3Event = record.getS3Event();
	switch(s3Event) {
		case ObjectRestoreCompleted:
			// do something
			break;
		case ObjectRestorePost:
			// do something else
	}
}
```  

Let me know what you think.
Thanks

dagnir:
Okay gotcha. In that case, I would rather add a method like `eventNameAsEnum` that does the `fromValue()`, but doesn't store a separate enum member. These objects should closely follow the service declared format of the object and adding the new member would violate that.

juanitosvq:
Ok I see where you are coming from. That makes sense to me, making the change now. Let me know if you have any other comments

dagnir:
Merged this upstream. It will be included in the next release. Thanks again!

Issue #2104, opened 2019-09-23
Step Function Builder is missing Map State


pgoranss:
The [Map State](https://docs.aws.amazon.com/step-functions/latest/dg/amazon-states-language-map-state.html) is missing from the Step Function Builder:

https://github.com/aws/aws-sdk-java/blob/master/aws-java-sdk-stepfunctions/src/main/java/com/amazonaws/services/stepfunctions/builder/StepFunctionBuilder.java

debora-ito:
Thank you for reporting this @pgoranss. 

Marking as a feature request but we don't have a date when this will be prioritized, as the team is actively working on V2 new features. PRs are welcome, and +1s help us spot the most requested features.

pgoranss:
@debora-ito PR submitted for this: https://github.com/aws/aws-sdk-java/pull/2113. Let me know if you need anything else.

debora-ito:
@pgoranss We added a task in our backlog to review the PR, will let you know when it's prioritized.

Stvad:
For people wondering about alternatives in the meantime - using AWS CDK can be an option https://docs.aws.amazon.com/cdk/api/latest/docs/aws-stepfunctions-readme.html

debora-ito:
Map State is now available from version `1.11.702`. Thank you!

PR #2113, opened 2019-10-04
Adding Map State to Step Functions Builder

pgoranss:
*Issue #, if available:*
https://github.com/aws/aws-sdk-java/issues/2104

*Description of changes:*
Adding the missing [MapState](https://docs.aws.amazon.com/step-functions/latest/dg/amazon-states-language-map-state.html) and corresponding Iterator to the Step Function Builder. This is, for the most part, following the same patterns used for the ParallelState Builder since the two are so similar. The main difference is rather than multiple branches, Map State contains a single _Iterator_ to apply against a list of inputs. 
There are also two additional fields for Map State that do not exist in Parallel State: 
**ItemsPath** - optional, follows the same pattern as InputPath, OutputPath, ResultPath
**MaxConcurrency** - optional, Integer

### DIFF
--- a/aws-java-sdk-stepfunctions/src/main/java/com/amazonaws/services/stepfunctions/builder/internal/validation/ValidationContext.java
+++ b/aws-java-sdk-stepfunctions/src/main/java/com/amazonaws/services/stepfunctions/builder/internal/validation/ValidationContext.java
@@ -159,6 +159,13 @@ final class ValidationContext {
     /**
      * Asserts that the string represents a valid JsonPath expression.
      *
+     * @param path Path expression to validate.
+     */
+    public void assertIsValidItemsPath(String path) { assertIsValidReferencePath(path, PropertyNames.ITEMS_PATH); }
+
+    /**
+     * Asserts that the string represents a valid JsonPath expression.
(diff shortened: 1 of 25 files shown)


aws-java-sdk-stepfunctions/src/main/java/com/amazonaws/services/stepfunctions/builder/internal/validation/ValidationContext.java
```diff
@@ -156,6 +156,13 @@ public void assertIsValidResultPath(String path) {
         assertIsValidReferencePath(path, PropertyNames.RESULT_PATH);
     }
 
+    /**
+     * Asserts that the string represents a valid JsonPath expression.
+     *
+     * @param path Path expression to validate.
+     */
+    public void assertIsValidItemsPath(String path) { assertIsValidReferencePath(path, PropertyNames.ITEMS_PATH); }
```
spfink: Separate this to multiple lines

spfink:
We've merged this through our internal release process and will go out publicly with tomorrow's release.

Thanks for the contribution!

PR #2121, opened 2019-10-15
Add public factory method getInstance to WebIdentityTokenCredentialsProvider

chancez:
The create() method does this, but other tools and libraries expect a
default constructor or getInstance() method for these purposes, such as
in the Hadoop s3a libraries.

For example if you attempt to use this provider in Hadoop explicitly:
```
<property>
  <name>fs.s3a.aws.credentials.provider</name>
  <value>com.amazonaws.auth.WebIdentityTokenCredentialsProvider</value>
</property>
```
Fails with:
```
Caused by: java.io.IOException: com.amazonaws.auth.WebIdentityTokenCredentialsProvider constructor exception.  A class specified in fs.s3a.aws.credentials.provider must provide a public constructor accepting Configuration, or a public factory method named getInstance that accepts no arguments, or a public default constructor.
        at org.apache.hadoop.fs.s3a.S3AUtils.createAWSCredentialProvider(S3AUtils.java:661) ~[hadoop-aws-3.1.1.jar:?]
        at org.apache.hadoop.fs.s3a.S3AUtils.cre
```
This PR should fix the above error.

### DIFF
--- a/aws-java-sdk-core/src/main/java/com/amazonaws/auth/WebIdentityTokenCredentialsProvider.java
+++ b/aws-java-sdk-core/src/main/java/com/amazonaws/auth/WebIdentityTokenCredentialsProvider.java
@@ -26,6 +26,12 @@ public class WebIdentityTokenCredentialsProvider implements AWSCredentialsProvid
     private final AWSCredentialsProvider credentialsProvider;
     private final RuntimeException loadException;
 
+    private static final WebIdentityTokenCredentialsProvider INSTANCE = new WebIdentityTokenCredentialsProvider();
+
+    private WebIdentityTokenCredentialsProvider() {
+        this(new BuilderImpl());
+    }
+
     private WebIdentityTokenCredentialsProvider(BuilderImpl builder) {
         AWSCredentialsProvider credentialsProvider = null;
         RuntimeException loadException = null;
@@ -75,6 +81,10 @@ public class WebIdentityTokenCredentialsProvider implements AWSCredentialsProvid
     @Override
     public void refresh() {
 
+    }
+
+    public static WebIdentityTokenCredentialsProvider getInstance() {
+        return INSTANCE;
     }
 
     public static WebIdentityTokenCredentialsProvider create() {


roitvt:
I use Spark on EKS and using [iam roles for service account](https://docs.aws.amazon.com/eks/latest/userguide/enable-iam-roles-for-service-accounts.html)
this PR is a must to make it work!

gnatpat:
I know it's bad practice to comment on MRs to say you also need them, but it looks like there's no reason why this shouldn't be merged - and as roitvt says, it would allow spark to use iam roles for service accounts!

bmaizels:
@chancez Thanks for making this pull request, looking at the other implementations of credentials providers in the same package, it looks like the most commonly employed pattern is to have a public default constructor. Is there a reason we should deviate from that pattern here?

chancez:
Honestly this is so old I have no idea what I was thinking when it was written.

seddonm1:
@chancez @bmaizels 
this PR came from this issue: https://github.com/kube-reporting/metering-operator/issues/947

This affects the `hadoop-aws` library which is depended on by Apache Hadoop and therefore Apache Spark (plus others) which needs to resolve a list of provider class names (from the Hadoop configuration) to provider instances:

https://github.com/apache/hadoop/blob/trunk/hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AUtils.java#L714

This PR adds the `getInstance` method so `com.amazonaws.auth.WebIdentityTokenCredentialsProvider` class can be instantiated by `hadoop-aws`
https://github.com/apache/hadoop/blob/trunk/hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AUtils.java#L744

zoewangg:
Hi all, we have made changes internally to add a public constructor in the `WebIdentityTokenCredentialsProvider` to be consistent with other provider classes. The change will be included in the next release.

Closing this PR. (@chancez Thank you for creating the PR!)

PR #2122, opened 2019-10-16
Update Jackson (jackson-databind) dependency to just released 2.6.7.3

cowtowncoder:
Update to `jackson-databind` dependency, from 2.6.7.2 to just released 2.6.7.3, which contains every security fix up to and including 2.9.10 (plus 2 fixes that will be in 2.9.10.1).
See https://github.com/FasterXML/jackson/wiki/Jackson-Release-2.6.7.x).
I realize that aws-sdk-java is not vulnerable (as per README), but many security tools can't tell the difference so this should be useful to many users.


cowtowncoder:
@spfink Makes sense. Happy to be able to help here!

bmaizels:
Change accepted and merged. Thank you very much for your contribution!

cowtowncoder:
Excellent, thanks!

PR #2158, opened 2019-11-22
com.fasterxml.jackson.core:jackson-databind version updated

AbtinZohrabi:
This package should be updated due to CVE-2019-14540


millems:
We're discovering that upgrading from 2.6.x to 2.10.x breaks a lot of our customers since they might be using features of Jackson that were removed before 2.10.x. We're currently exploring alternatives, but we're not able to merge this PR right now.

You can upgrade Jackson on your end until we figure out a workaround on ours, assuming your application supports 2.10.x. The SDK itself is forward-compatible with 2.10.x.

PR #2166, opened 2019-12-06
GetObjectRequest should support final bytes as a Range header value

### LINKED ISSUE #1551: GetObjectRequest in S3 should support final bytes as a Range header value
According to https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.35, a Range header may include a single negative value to indicate the last X bytes in a file should be retrieved. For example `bytes=-500` is a valid Range value for the final 500 bytes in a file.

This Range header option is currently supported by S3, as verified through the AWS S3 CLI.

Currently, GetObjectRequest includes `setRange(long start)` and `setRange(long start, long end)`, which supports Range values like `bytes=100-` and `bytes=100-200`, however, there is no way to provide a Range value in GetObjectRequest which results in "bytes=-100", despite the fact that this is a valid value which is already supported by S3.

omalley:
Fixes #1551

*Issue #, if available:* 1551

*Description of changes:*

Adds support for ranges that get the tail of objects.

### DIFF
--- a/aws-java-sdk-core/src/main/java/com/amazonaws/AmazonWebServiceClient.java
+++ b/aws-java-sdk-core/src/main/java/com/amazonaws/AmazonWebServiceClient.java
@@ -179,13 +179,14 @@ public abstract class AmazonWebServiceClient {
      */
     public AmazonWebServiceClient(ClientConfiguration clientConfiguration,
             RequestMetricCollector requestMetricCollector) {
-        this(clientConfiguration, requestMetricCollector, false);
+        this(clientConfiguration, requestMetricCollector, false, null);
     }
 
     @SdkProtectedApi
     protected AmazonWebServiceClient(final ClientConfiguration clientConfiguration,
                                      final RequestMetricCollector requestMetricCollector,
-                                     boolean disableStrictHostNameVerification) {
+                                     boolean disableStrictHostNameVerification,
+                                     final AmazonHttpClient httpClient) {
         this(new AwsSyncClientParams() {
             @Override
             public AWSCredentialsProvider getCredentialsProvider() {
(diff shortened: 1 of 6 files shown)


millems:
Would you mind commenting with a few tests for this change? There's no place to import them in github for the V1 SDK, but we execute them internally, so having some that we can copy would help a lot.

Make sure to redact any credentials if you write integration tests.

omalley:
It seemed better to make a real unit test for the change, although I had a to add a parameter so that I could get a mock http client to the s3 client. With that change, my test is able to change the range parameters and validate that the proper headers are being given to the http client.

I obviously did also write some code that uses the patch to access a real s3 object if you want that as well, but it isn't very interesting code.

millems:
@omalley Awesome! I'd love the uninteresting code as well to copy+paste into our regression test suite.

omalley:
The function that I wrote takes three parameters:
- the bucket name
- the key 
- the offset to read from (with negative values using the new code path)

```import com.amazonaws.services.s3.AmazonS3;
import com.amazonaws.services.s3.AmazonS3ClientBuilder;
import com.amazonaws.services.s3.model.GetObjectRequest;
import com.amazonaws.services.s3.model.ObjectMetadata;
import com.amazonaws.services.s3.model.S3Object;
import com.amazonaws.services.s3.model.S3ObjectInputStream;

public class S3Test {
    public static void main(String[] args) throws Exception {
        AmazonS3 client = AmazonS3ClientBuilder.standard().withRegion("us-west-2").build();
        S3Object obj = client.getObject(new GetObjectRequest(args[0], args[1])
                                            .withRange(Long.parseLong(args[2])));
        S3ObjectInputStream stream = obj.getObjectContent();
        byte[] buffer = new byte[25];
        long total = 0;
        int len = stream.read(buffer);
        while (len != -1) {
            total += len;
            StringBuilder line = new StringBuilder();
            for(int i=0; i < len; ++i) {
                if (i != 0) {
                    line.append(' ');
                }
                line.append(String.format("%02x", buffer[i] & 0xff));
            }
            System.out.println(line);
            len = stream.read(buffer);
        }
        stream.close();
        ObjectMetadata meta = obj.getObjectMetadata();
        if (meta != null) {
            Long[] range = meta.getContentRange();
            if (range != null) {
                System.out.println("Range="+range[0]+","+range[1]);
            }
        }
        obj.close();
        System.out.println("Length = " + total);
    }
}
```

cenedhryn:
Status update: We have assigned a rough priority to incorporating this PR into the SDK and expect to work on it in February, pending other work and events.

dagnir:
The change to `AmazonS3Client` itself looks fine to me, thanks!

I want to avoid pulling in a new dependency (EasyMock), so I'll convert the tests here to a format more in line with our existing tests when I pull it in upstream.

PR #2174, opened 2019-12-09
Use configured retry policy for CompleteMultipartUpload

### LINKED ISSUE #2168: CompleteMultipartUpload is not retried on throttling
CompleteMultipartUpload doesn't use provided retry policy, instead it relies on internal `CompleteMultipartUploadRetryCondition` https://github.com/aws/aws-sdk-java/blob/master/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/AmazonS3Client.java#L465-L466

This condition uses a static predicate that only checks for a single error: https://github.com/aws/aws-sdk-java/blob/master/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/internal/CompleteMultipartUploadRetryablePredicate.java#L48-L49:
- error code: `InternalError`
- error message should contain `Please try again.`

In case of throttling, the error code is actually `SlowDown` and the error message is `Please reduce your request rate.`, which means the predicate is always false and the `CompleteMultipartUpload` is never retried on throttling. 

Documention states it's safe to retry: https://docs.aws.amazon.com/AmazonS3/latest/API/API_CompleteMultipartUpload.html 

**Potential solutions**:
- Extend `CompleteMultipartUploadRetryablePredicate` to support more errors
- Use a configured retry policy instead of `CompleteMultipartUploadRetryCondition`

sap1ens:
Closes #2168

Using a default configured retry policy inside the `shouldRetryCompleteMultipartUpload` method. This makes sure all potential errors (including throttling) are handled. `CompleteMultipartUploadRetryCondition` only handles one specific error.


sap1ens:
Could someone point me to the relevant tests? I feel silly, but I couldn't find any for the S3 client.

millems:
Sorry, the tests are kept internal in V1 of the Java SDK. I know that makes contributions difficult.

We'll take a look and get back to you on this. I'm a little nervous as to why we had it this way to begin with.

sap1ens:
Thanks! 

FYI, this has been running in production on our side for about a week, all good so far.

debora-ito:
Closing this after discussion in #2168.

Issue #2176, opened 2019-12-10
ClientConfiguration should ignore empty proxy env variables


pedrolauro:
Although my proxy env variables are all set to empty, SDK keeps trying to parse that:

```
[WARN ] 2019-12-10 12:11:11.068-0200 [restartedMain] ClientConfiguration - Unable to parse HTTPS_PROXY environment variable value '' as URL. It is malformed.
java.net.MalformedURLException: no protocol: 
	at java.net.URL.<init>(URL.java:593) ~[?:1.8.0_191]
	at java.net.URL.<init>(URL.java:490) ~[?:1.8.0_191]
	at java.net.URL.<init>(URL.java:439) ~[?:1.8.0_191]
	at com.amazonaws.ClientConfiguration.getUrlEnvVar(ClientConfiguration.java:2449) ~[aws-java-sdk-core-1.11.688.jar:?]
	at com.amazonaws.ClientConfiguration.getHttpProxyEnvironmentVariable(ClientConfiguration.java:2440) ~[aws-java-sdk-core-1.11.688.jar:?]
	at com.amazonaws.ClientConfiguration.getProxyHostEnvironment(ClientConfiguration.java:700) ~[aws-java-sdk-core-1.11.688.jar:?]
	at com.amazonaws.ClientConfiguration.getProxyHost(ClientConfiguration.java:728) ~[aws-java-sdk-core-1.11.688.jar:?]
	at com.amazonaws.http.settings.HttpClientSettings.getProxyHost(HttpClientSettings.java:84) ~[aws-java-sdk-core-1.11.688.jar:?]
	at com.amazonaws.http.settings.HttpClientSettings.isProxyEnabled(HttpClientSettings.java:176) ~[aws-java-sdk-core-1.11.688.jar:?]
	at com.amazonaws.http.apache.client.impl.ApacheHttpClientFactory.addProxyConfig(ApacheHttpClientFactory.java:88) ~[aws-java-sdk-core-1.11.688.jar:?]

...
```

My suggestion is to change method `ClientConfiguration.getUrlEnvVar` to check not only for null values, but also for empty ones:

```
package com.amazonaws;
...
public class ClientConfiguration {
....

private URL getUrlEnvVar(AtomicReference<URLHolder> cache, String name) {
        if (cache.get() == null) {
            URLHolder holder = new URLHolder();
            String value = getEnvironmentVariableCaseInsensitive(name);
            // if (value != null) { -- old version
[...]

PR #2178, opened 2019-12-11
Empty environment variables should be treated as null when dealing with proxy configuraton

pedrolauro:
Change on `getEnvironmentVariable` method to check for empty variables, and return those as null. This will evict malformed URL warnings when there's actually no proxy configured. Also some tests to check that new condition.

Issue #2176


pedrolauro:
Good morning, @KaibaLopez. There's the PR. Could you review it?

millems:
@pedrolauro I can review. Thanks for the PR!

millems:
I'll merge in the next day or so. We have to mirror the change commit to our internal master repository so that our automation doesn't override it. Thanks for your contribution!

pedrolauro:
> I'll merge in the next day or so. We have to mirror the change commit to our internal master repository so that our automation doesn't override it. Thanks for your contribution!

Ok. You're welcome. I feel glad to help.

PR #2223, opened 2020-02-12
updated plugins for Reproducible Builds (excl osgi)

hboutemy:
aws-java-sdk-osgi is the only remaining non-reproducible module, waiting
for Felix maven-bundle-plugin improvement

see https://maven.apache.org/guides/mini/guide-reproducible-builds.html for more information on how to configure a Maven build to get Reproducible Build

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #2244, opened 2020-03-01
Add Build-Time to jar MANIFEST

khansuhel:
*Issue #, if available:* NA

*Description of changes:* My organization uses a tool that keeps the dependencies updated. It does so by reading the version of the dependencies mentioned in pom.xml and checking if there is an update available. However, its not very wise to update the libraries as soon as they are released and our tool provides a way to configure the number of days it should wait after a library is released. And for this it looks for `Build-Time` in jar's MANIFEST.MF file. This PR aims to achieve the same.

Useful link: https://maven.apache.org/plugin-developers/cookbook/add-build-time-to-manifest.html

### DIFF
--- a/aws-java-sdk-core/pom.xml
+++ b/aws-java-sdk-core/pom.xml
@@ -102,11 +102,53 @@
       <scope>test</scope>
     </dependency>
   </dependencies>
+
   <build>
+    <resources>
+      <resource>
+        <directory>src/main/resources</directory>
+        <filtering>true</filtering>
+      </resource>
+    </resources>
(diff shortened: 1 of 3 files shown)


hboutemy:
please don't do that: this will go against Reproducible Builds principles
see #2223

millems:
It seems like this is a potentially-controversial request, because of its impact on build reproducibility. I feel like the right solution is to set the build-time value to a static value for non-SNAPSHOT versions. That might be something we could do in V2, but we're not willing to invest the time into for V1.

Issue #2180, opened 2019-12-16
Unable to use KVS PutMediaClient behind proxy


luclement:
Hi,

I need to call the `PutMedia` operation for Kinesis Video Streams behind a corporate proxy. Unfortunately, since the `AmazonKinesisVideoPutMediaClientBuilder` does not extend the `AwsClientBuilder` there is no way to set the `ClientConfiguration`.

Additionally, setting Java/system environment variables doesn't seem to work either because the PutMedia client uses Netty as transport and would, based on some research, require adding a proxy handler first in the channel pipeline.

https://github.com/aws/aws-sdk-java/blob/89fd27ef2abd407ad64245215f9e90dd40276651/aws-java-sdk-kinesisvideo/src/main/java/com/amazonaws/services/kinesisvideo/internal/netty/handler/PutMediaHandlerInitializer.java#L48

Let me know if I'm seeing this correctly. 

Thanks.

luclement:
FYI I got this to work by adding an `HttpProxyHandler` as the first step in the `ChannelPipeline`

debora-ito:
Echoing the comments in the PR, we'll not implement this feature in v1 as the team's efforts are focused in the SDK v2.
Tracking issue for the implementation of the equivalent functionality to AmazonKinesisVideoPutMedia client in v2: https://github.com/aws/aws-sdk-java-v2/issues/1330.

PR #2246, opened 2020-03-02
Add proxy options for KVS PutMediaClient

luclement:
*Issue #, if available:*
#2180 

*Description of changes:*
This PR adds optional `.withProxyHost` and `.withProxyPort` configurations while creating a `AmazonKinesisVideoPutMediaClient`. This adds an additional Netty `ProxyHandler` as the first step in the channel pipeline. If either `proxyHost` or `proxyPort` are null, i.e. not configured, the proxy handler will not be added to the pipeline mimicking the original channel behavior.

If the proxy handler is defined, we set the resolver as such: `b.resolver(NoopAddressResolverGroup.INSTANCE)`. This ensures that Netty does not resolve the target hostname before the initial CONNECT to prevent breaking any proxy hostname whitelist conditions. See: https://github.com/netty/netty/issues/4526

I didn't see an existing test suite for any of the `kinesisvideo` stuff. Can work on adding one if needed. The client should continue to work as is if the proxy port and host aren't configured; otherwise, they can be configured using the builder.


millems:
Unfortunately we can't add new dependencies very easily. For the SDK V2 we implemented this stuff ourselves: https://github.com/aws/aws-sdk-java-v2/commit/039d98f9e85e892158d21902e621524704598d7a#diff-fcaf9f532153a7d4f6e802677b58fb82

We're also hesitant to implement new features in V1 of the SDK, because such features keep pulling us off of V2, both to review them and to fix bugs that customers report afterwards.

For both of these reasons, we're not able to take this pull request. Sorry about the inconvenience. When we implement this feature in V2, we should be able to support proxies with the put-media client: https://github.com/aws/aws-sdk-java-v2/issues/1330

Issue #2263, opened 2020-03-15
TransferManager multipart parellel download failures not forwarded back to the client


jamesceddy:
I am using TransferManager to download files on an unstable connection. It is not unusual to see read timeouts on connections, so failures are expected. However, TransferManager is leaving Downloads in "InProgress" states indefinatly. Because of this, I have no way to know if TransferManager is finished or not. If TransferManager is done, it should report a terminal state back to the progress listeners.

This  issue can be seen with these steps
(1) start N multipart downloads.
(2) use iptables -J DROP to simulate network loss
(3) wait for the outcome from the ProgressListeners

Expected: There is some outcome, once the downloads are completed they should be in any of the terminal states: success/failure/cancel.
Actual: Callbacks to progress listeners are in a non-terminal state (InProgress) in-defiantly.

```
import com.amazonaws.ClientConfiguration
import com.amazonaws.auth.AWSCredentials
import com.amazonaws.auth.AWSCredentialsProvider
import com.amazonaws.auth.BasicAWSCredentials
import com.amazonaws.event.ProgressEventType
import com.amazonaws.event.ProgressListener
import com.amazonaws.retry.PredefinedRetryPolicies
import com.amazonaws.retry.RetryMode
import com.amazonaws.services.s3.AmazonS3ClientBuilder
import com.amazonaws.services.s3.transfer.Download
import com.amazonaws.services.s3.transfer.TransferManagerBuilder
import java.io.File
import java.util.concurrent.*
import java.util.concurrent.atomic.AtomicBoolean
import kotlin.concurrent.thread


val downloads = mutableMapOf<String, Download>()
val downloadLastCallbackEvents = ConcurrentHashMap<String, ProgressEventType>()
val isNetworkBlockerStarted = AtomicBoolean(false)
val fileKeys = mutableListOf(
    "key/file42mb",
    "key/file22mb",
    "key/file18mb",
    "key/file10mb"
)
fun main() {
    netUp()
    val accessKey = "zzzz"
    val secretKey = "xyz"

    val s3Client = AmazonS3ClientBuilder.standard().withRegion("us-east-1").withCredentials(object :AWSCredentialsProvider{
        override fun getCredentials(): AWSCredentials {
            return BasicAWSCredentials(accessKey, secretKey)
        }

        override fun refresh() {
        }
    }).build()
[...]

PR #2265, opened 2020-03-15
Report correct transfer state on downloads when parts timeout

jamesceddy:
When parts are downloading exceptions thrown from socket timeouts can cause download states to be reported as "InProgress" indefinitely.

fixes: https://github.com/aws/aws-sdk-java/issues/2263

### DIFF
--- a/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/transfer/internal/CompleteMultipartDownload.java
+++ b/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/transfer/internal/CompleteMultipartDownload.java
@@ -14,12 +14,14 @@
  */
 package com.amazonaws.services.s3.transfer.internal;
 
+import com.amazonaws.SdkClientException;
 import com.amazonaws.annotation.SdkInternalApi;
 import com.amazonaws.services.s3.internal.FileLocks;
 import com.amazonaws.services.s3.transfer.Transfer;
 import java.io.File;
 import java.util.List;
 import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Future;
 
 /**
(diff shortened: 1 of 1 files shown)


aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/transfer/internal/CompleteMultipartDownload.java
```diff
@@ -41,13 +43,23 @@ public CompleteMultipartDownload(List<Future<Long>> files, File destinationFile,
...
-                long filePosition = file.get();
+            for (; index < partFiles.size(); index++) {
+                long filePosition = partFiles.get(index).get();
                 download.updatePersistableTransfer(currentPartNumber++, filePosition);
             }
-
             download.setState(Transfer.TransferState.Completed);
+        } catch (ExecutionException e) {
+            // if any part fails, we cancel remaining part downloads and notify clients of the failure.
+            for (; index < partFiles.size(); index++){
```
suthagar23: Based on this implementation, all the uploads next to the exception-upload will be canceled (including the upload which got exception). So what about the uploads happened before to this exception (which are still pending at this point)?

jamesceddy: Since each individual piece is a future to an executor, which may or may not be a single executor, you can fail out of order. For example, you can have this failure 

Blocks => D = downloaded, f = failure, p = pending

**Ex1:**
file = { D, D, D, f, p , p, D}

But you cannot have this failure:
**Ex2:**
file = {D, p, D, f, _, _, _, _}

or this failure:
**Ex3:**
file = {D, f, D, f, _, _, _, _}


In "Ex1" your file wrote 3 blocks, has 2 blocks pending and 1 block written at the end that's good. You'll lose that block, but you would anyway if you paused the download at that point.

"Ex2" and "Ex3" cannot exist because CompletableDownload forces assembly in serial order.

The PersitableDownload feature, and the CompletableDownload are both based on serial order. That's a constraint within the existing implementation. 

When you say "uploads" it's a little bit ambiguous to me if you are referring to a "part upload" or a "file upload". To try to make this clear, the CompletableDownload is a container for a single file, and each future is a part of that file. Which means (as explained above), on failure, you'll get the a continuous sequence of completed parts from the "start" to block N. Which will allow the caller to, if they so chose, get a handle to the PersitableDownload at which they can resume at a later time.

Hope it helps.

sidkhangit:
I see this changes are not present in master branch, when they will get merged to master branch?

PR #2266, opened 2020-03-15
When multipart uploads fail, cancel remaining part transfers

jamesceddy:
Suppose we have 10 parts, if we fail on part #2 we fire a "TRANSFER_FAILED_EVENT" back to the client. We should assume that clients will react to this failure, possibly with custom logic to retry. Supposing that part #3 succeeds, the client will then get other transfer events past the "TRANSFER_FAILED_EVENT". I assume here that "TRANSFER_FAILED_EVENT" is a terminal event and that a client should not expect to receive further events for that transfer, successful or not. However, the meaning of TRANSFER_FAILED_EVENT is not clearly defined in the javadocs.

relates to: https://github.com/aws/aws-sdk-java/issues/2267

### DIFF
--- a/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/transfer/internal/CompleteMultipartUpload.java
+++ b/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/transfer/internal/CompleteMultipartUpload.java
@@ -115,17 +115,23 @@ public class CompleteMultipartUpload implements Callable<UploadResult> {
      */
     private List<PartETag> collectPartETags() {
 
-        final List<PartETag> partETags = new ArrayList<PartETag>();
+        final List<PartETag> partETags = new ArrayList<PartETag>(eTagsBeforeResume.size());
         partETags.addAll(eTagsBeforeResume);
-        for (Future<PartETag> future : futures) {
-            try {
-                partETags.add(future.get());
-            } catch (Exception e) {
-                throw new SdkClientException(
-                        "Unable to complete multi-part upload. Individual part upload failed : "
-                                + e.getCause().getMessage(), e.getCause());
+
+        int index = 0;
+        try {
+            for (; index < futures.size(); index++) {
+                partETags.add(futures.get(index).get());
             }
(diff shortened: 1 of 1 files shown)


aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/transfer/internal/CompleteMultipartUpload.java
```diff
@@ -115,17 +115,23 @@ public UploadResult call() throws Exception {
      */
     private List<PartETag> collectPartETags() {
 
-        final List<PartETag> partETags = new ArrayList<PartETag>();
+        final List<PartETag> partETags = new ArrayList<PartETag>(eTagsBeforeResume.size());
```
suthagar23: What do you think about, 
```suggestion
        final List<PartETag> partETags = new ArrayList<PartETag>(eTagsBeforeResume.size() +  futures.size());
```

aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/transfer/internal/CompleteMultipartUpload.java
```diff
@@ -115,17 +115,23 @@ public UploadResult call() throws Exception {
...
-                        "Unable to complete multi-part upload. Individual part upload failed : "
-                                + e.getCause().getMessage(), e.getCause());
+
+        int index = 0;
+        try {
+            for (; index < futures.size(); index++) {
+                partETags.add(futures.get(index).get());
             }
+        } catch (Exception e) {
+            for (; index < futures.size(); index++) {
```
suthagar23: Don't we need to cancel all the uploads? 
Based on this implementation, all the uploads next to the exception-upload will be canceled (including the upload which got exception). So what about the uploads happened before to this exception?

zoewangg:
Thank you again for your PR! We have made the change internally and it should be available in the next release.

sidkhangit:
I see this changes are not present in master branch, when they will get merged to master branch?

Issue #2274, opened 2020-03-23
com.amazonaws.regions.EndpointToRegion#guessRegionNameForEndpoint Broken for Endpoint String without Schema


original-brownbear:
After upgrading to `1.11.749`, I see this failure when constructing an S3 client that uses e.g. endpoint string `127.0.0.1:5021` or any other string that doesn't contain a schema.

```
  Caused by: java.lang.IllegalArgumentException: Illegal character in scheme name at index 0: 127.0.0.1:5
  1>    at java.base/java.net.URI.create(URI.java:906)
  1>    at com.amazonaws.regions.EndpointToRegion.guessRegionOrRegionNameForEndpoint(EndpointToRegion.java:71)
  1>    at com.amazonaws.regions.EndpointToRegion.guessRegionNameForEndpoint(EndpointToRegion.java:41)
  1>    at com.amazonaws.AmazonWebServiceClient.setEndpoint(AmazonWebServiceClient.java:323)
  1>    at com.amazonaws.services.s3.AmazonS3Client.setEndpoint(AmazonS3Client.java:725)
  1>    at com.amazonaws.client.builder.AwsClientBuilder.setRegion(AwsClientBuilder.java:453)
  1>    at com.amazonaws.client.builder.AwsClientBuilder.configureMutableProperties(AwsClientBuilder.java:424)
  1>    at com.amazonaws.client.builder.AwsSyncClientBuilder.build(AwsSyncClientBuilder.java:46)
```

The JavaDocs explicitly state that an endpoint string without a schema should work and it used to work in `1.11.636` so this is new it seems.
I think this can easily be fixed by using e.g. `com.amazonaws.AmazonWebServiceClient#toURI` which is used in other places to handle this kind of string to `URI` conversion.

debora-ito:
Hi @original-brownbear, thank you for reaching out. This is a duplicate of #2255, let's track this issue over there.

Issue #2255, opened 2020-03-06
IP address & port w/o scheme in endpoint ⇒ URISyntaxException/IllegalArgumentException from EndpointToRegion.guessRegionOrRegionNameForEndpoint


jglick:
## Describe the bug

If you pass a custom endpoint in the format `127.0.0.1:1234` to the SDK, the client builder will fail.

## Expected Behavior

Should behave similarly to if a hostname such as `localhost` is used, or if an IP address is used without an explicit port, or if a full URL with scheme is passed.

## Current Behavior

You get an error such as

```
java.lang.IllegalArgumentException: Illegal character in scheme name at index 0: 127.0.0.1:38835
	at java.base/java.net.URI.create(URI.java:883)
	at com.amazonaws.regions.EndpointToRegion.guessRegionOrRegionNameForEndpoint(EndpointToRegion.java:71)
	at com.amazonaws.regions.EndpointToRegion.guessRegionNameForEndpoint(EndpointToRegion.java:41)
	at com.amazonaws.AmazonWebServiceClient.setEndpoint(AmazonWebServiceClient.java:323)
	at com.amazonaws.services.s3.AmazonS3Client.setEndpoint(AmazonS3Client.java:725)
	at com.amazonaws.client.builder.AwsClientBuilder.setRegion(AwsClientBuilder.java:453)
	at com.amazonaws.client.builder.AwsClientBuilder.configureMutableProperties(AwsClientBuilder.java:424)
	at com.amazonaws.client.builder.AwsSyncClientBuilder.build(AwsSyncClientBuilder.java:46)
	at …
Caused by: java.net.URISyntaxException: Illegal character in scheme name at index 0: 127.0.0.1:38835
	at java.base/java.net.URI$Parser.fail(URI.java:2913)
	at java.base/java.net.URI$Parser.checkChars(URI.java:3084)
	at java.base/java.net.URI$Parser.checkChar(URI.java:3094)
	at java.base/java.net.URI$Parser.parse(URI.java:3109)
	at java.base/java.net.URI.<init>(URI.java:600)
	at java.base/java.net.URI.create(URI.java:881)
	... 23 more
```

## Steps to Reproduce

Reproduced in https://github.com/jenkinsci/artifact-manager-s3-plugin/pull/112.

## Possible Solution

PR forthcoming.

## Context

Broke integration testing of the Jenkins plugin to store artifacts in S3 against a MinIO mock server running in a local Docker container.
[...]

PR #2277, opened 2020-03-25
fix for #2274/ #2255 EndpointToRegion-> ip adress & port without sche…

TommyWind:
…me in endpoint

*Issue #2274/ #2255 

*Description of changes: fixes EndpointToRegion#guessRegionNameForEndpoint for Endpoint String without Schema. Example added to tests. 

### DIFF
--- a/aws-java-sdk-core/src/main/java/com/amazonaws/regions/EndpointToRegion.java
+++ b/aws-java-sdk-core/src/main/java/com/amazonaws/regions/EndpointToRegion.java
@@ -67,12 +67,10 @@ public class EndpointToRegion {
         if (endpoint == null) {
             return new RegionOrRegionName();
         }
-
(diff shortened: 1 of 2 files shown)


TommyWind:
@debora-ito I've just seen there is an opened PR for the same issue. Sorry for duplication. If it is ok, I am going to leave it here for a review anyway. This solution looks for me a bit leaner. Thank you in advance.🙂

zoewangg:
Hi @TommyWind Thank you for your pull request! After careful reviewing the code, we have decided to merge #2256 instead because there might be some other edge cases that cause `URI.create` to throw exception such as  `://127.0.0.1`. We think catching exception might be a better solution for this case. Going to go ahead and close this PR. Thank you again for the pull request! :)

PR #2334, opened 2020-05-23
Correct javadoc ending

asashour:
This PR corrects the endings of the JavaDocs.

Also in template files


debora-ito:
Hi @asashour I applaud the effort of changing 696 files but these classes are generated, the change would be overwritten by the next release, so we cannot approve this change unfortunately.

asashour:
I understand that the files are generated from the templates, which are also included in the PR.

Should I raise a new PR with only template changes?

Please note there are other issues with the javadoc, like using tabs instead of spaces and incorrect indentation. Which should be tackled by other PRs once the change process is agreed on

debora-ito:
You can submit a new PR for just the templates, it would not be high in the priority list as the team is actively working in releasing new features for the SDK v2.

Out of curiosity, is this impacting you in some way? Can you give context on the change?

PR #2354, opened 2020-06-07
Improve grammar in documentation (Values values to Values range).

Isabek:
*Issue #, if available:*

*Description of changes:*

Change Values values to Values range.


debora-ito:
Hi @Isabek this javadoc is generated based on the service model, which is owned by the service team. I'll contact them and ask to make this improvement. Thank you for reaching out.

PR #2364, opened 2020-06-16
 Update ObjectMetadata getContentLength javadoc to correctly describe the default value

emmberk:
*Issue #, if available:*
none

*Description of changes:*
Fixes the javadoc of `ObjectMetadata#getContentLength()` to say that it returns `0` instead of `null` "if [the Content-Length header] hasn't been yet" to match the implementation of the method, which does
```java
if (contentLength == null) return 0;
```


emmberk:
This is still an issue in the latest release: https://github.com/aws/aws-sdk-java/blob/1.12.490/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/model/ObjectMetadata.java#L296-L298.

debora-ito:
@emmberk we incorporated the change into our internal code, it was released yesterday -
https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/s3/model/ObjectMetadata.html#getContentLength--

Thank you for the patience! We appreciate the persistence.

PR #2380, opened 2020-07-14
support weak etags in getObject checks

tkroman:
According to ETag spec (https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag) ETags can be of weak form, which is the case for us when we use Ceph as a storage. In general, I don't think this is backwards-incompatible as no existing functionality is harmed. 
With this change support for etags marked as weak is added.
Not sure what's the QA process is supposed to be but I guess a brief glance at the code should suffice.

### DIFF
--- a/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/AmazonS3Client.java
+++ b/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/AmazonS3Client.java
@@ -1544,7 +1544,12 @@ public class AmazonS3Client extends AmazonWebServiceClient implements AmazonS3 {
         // stream in a validator that calculates an MD5 of the downloaded
         // bytes and complains if what we received doesn't match the Etag.
         if (!skipClientSideValidation) {
-            byte[] serverSideHash = BinaryUtils.fromHex(s3Object.getObjectMetadata().getETag());
+            String etag = s3Object.getObjectMetadata().getETag();
+            if (etag.startsWith("W/")) {
+                // weak etags are of the form `W/"etag"` so we trim accordingly
+                etag = etag.substring(3, etag.length() - 1);
+            }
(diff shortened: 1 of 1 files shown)


debora-ito:
Hi @tkroman I am not sure if S3 supports weak etags.
Can you give more context and some examples on how this would work with Ceph?

tkroman:
Short example:

```
@ s3.getObject(bucket, key)
java.lang.IllegalArgumentException: Input is expected to be encoded in multiple of 2 bytes but found: 35
  com.amazonaws.util.Base16Codec.decode(Base16Codec.java:76)
  com.amazonaws.util.Base16Lower.decode(Base16Lower.java:53)
  com.amazonaws.util.BinaryUtils.fromHex(BinaryUtils.java:48)
  com.amazonaws.services.s3.AmazonS3Client.postProcessS3Object(AmazonS3Client.java:1547)
  com.amazonaws.services.s3.AmazonS3Client.getObject(AmazonS3Client.java:1502)
  com.amazonaws.services.s3.AmazonS3Client.getObject(AmazonS3Client.java:1344)
  ammonite.$sess.cmd36$.<clinit>(cmd36.sc:1)
```

BUT

```
@ s3.getObjectMetadata(bucket, key).getETag
res37: String = "09757082cae2ec0f59688f3b2f964e00"
```

I debugged the `postProcessS3Object` and found that somehow in transit the metadata on the object received by `postProcessS3Object` contains etag in `W/"..."` format. I can see that the metadata population in both cases is done in the same place (`AbstractS3ResponseHandler#populateObjectMetadata`) but I can't tell how it ends up having different behavior.

debora-ito:
Apologies for the long silence. We are closing this PR, we don't have plans to change the S3 etag handling. We are reserving the changes in the S3 handwritten client to the minimum necessary.

Issue #2384, opened 2020-07-17
Include DeliveryDelay in the SES EventType enum


alexburlton-sonocent:
## Describe the issue
The new event type `DeliveryDelay` was [recently added to SES](https://aws.amazon.com/about-aws/whats-new/2020/06/amazon-ses-can-now-send-notifications-when-the-delivery-of-an-email-is-delayed/), however the SDK has not been updated to include it in the `EventType` enum. It would be useful to update this as I imagine anyone who is doing any processing of these events will likely be making use of it in parsing/handling logic.

## Your Environment

* AWS Java SDK version used: 1.11.630

debora-ito:
Hi @alexburlton-sonocent thank you for reaching out. 

According to [SES docs](https://docs.aws.amazon.com/ses/latest/DeveloperGuide/monitor-using-event-publishing.html#event-publishing-terminology), the DELIVERY_DELAY type is only available via SES API v2 `UpdateConfigurationSetEventDestination` operation, which uses the SES API v2 [EventType](https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/simpleemailv2/model/EventType.html) and already has the DELIVERY_DELAY included.

PR #2385, opened 2020-07-17
Add DeliveryDelay to SES EventType enum

alexburlton-sonocent:
*Issue #, if available:*

https://github.com/aws/aws-sdk-java/issues/2384

*Description of changes:*

Manually adds the `DeliveryDelay` event type to the relevant `enum` class. This may not be the right thing to do - I am aware that this class appears to have been automatically generated somehow (`@Generated("com.amazonaws:aws-java-sdk-code-generator")`). However, as there is no `README` in the code generator stuff I've got no idea how I'd go about rerunning/changing that so I thought I'd see what the feedback was :shrug: 


debora-ito:
As I mentioned in #2384 the SES API v2 EventType has DELIVERY_DELAY included, so this change is not necessary.

PR #2424, opened 2020-10-03
support for client generation using credentials

15pratik:
*Issue #, if available:*

*Description of changes:*
Added an API for s3 client generation using access key and secret key without using the profile keys


debora-ito:
@15pratik we moved away from your suggested pattern of providing arguments. Thank you for the PR but we won't be able to approve it.

PR #2428, opened 2020-10-07
Change changelog of version 1.11.618 to add the docs of STS regional endpoints setting

Quanzzzz:
*Description of changes:*
We've enabled the setting of STS regional endpoints option, but this change wasn't added in the changelog. To prevent any misunderstanding about this behavior, and let the customers track this change easilier, here we add a new item in the change log about this change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4950,6 +4950,10 @@
   - ### Features
     - Amazon Transcribe - support transcriptions from audio sources in Russian (ru-RU) and Chinese (zh-CN).
 
+## __AWS SDK for Java__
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -4950,6 +4950,10 @@
   - ### Features
     - Amazon Transcribe - support transcriptions from audio sources in Russian (ru-RU) and Chinese (zh-CN).
 
+## __AWS SDK for Java__
```
dagnir, line 4953: You can't edit this file directly because it will be overwritten when the new is generated.

PR #2438, opened 2020-10-19
Better customization of DefaultBatchLoadRetryStrategy

ZeldaZach:
*Issue #, if available:*
N/A

*Description of changes:*
The `DefaultBatchLoadRetryStrategy` did not support customization for the number of retries to attempt. I found this odd as the `DefaultBatchWriteRetryStrategy` did allow this feature, so I ported a fix over. While a small change, I find this to be a net positive as these operations are very similar and having different ctor args felt off to me.

I'm a SDE within Prime Video and this use case came up for us when we wanted to specify our own Builder config. I find it interesting how we're the first group to want a change to the default max retry count.


debora-ito:
@ZeldaZach apologies for taking this long to reply.

We don't have plans to add significant features to the Java SDK v1. I'll close this as stale.

PR #2446, opened 2020-11-05
Fix document link

lhm7877:
Fix document link


debora-ito:
@lhm7877 thank you for the PR. The change was merged internally (finally!) and will be available in the next release.

PR #2454, opened 2020-11-10
Update STSAssumeRoleWithWebIdentitySessionCredentialsProvider.java

jeff-miller-cfa:
This is to fix a problem where a InvalidIdentityTokenException does not actually trigger the retries.

When the AmazonHttpClient creates the exception there is no inner cause so comparing the inner cause with the instance type doesn't work.

### DIFF
--- a/aws-java-sdk-sts/pom.xml
+++ b/aws-java-sdk-sts/pom.xml
@@ -33,6 +33,12 @@
         <groupId>com.amazonaws</groupId>
         <optional>false</optional>
     </dependency>
+    <dependency>
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+      <version>${junit.version}</version>
+      <scope>test</scope>
+    </dependency>
 </dependencies>
 
   <build>
--- a/aws-java-sdk-sts/src/main/java/com/amazonaws/auth/STSAssumeRoleWithWebIdentitySessionCredentialsProvider.java
+++ b/aws-java-sdk-sts/src/main/java/com/amazonaws/auth/STSAssumeRoleWithWebIdentitySessionCredentialsProvider.java
@@ -227,7 +227,7 @@ public class STSAssumeRoleWithWebIdentitySessionCredentialsProvider implements A
         }
     }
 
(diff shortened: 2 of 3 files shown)


Quanzzzz:
Hi @jeff-miller-cfa, thanks for your pr, this change is reasonable. But could you include corresponding tests with the change so we can make sure it's working?

jeff-miller-cfa:
> Hi @jeff-miller-cfa, thanks for your pr, this change is reasonable. But could you include corresponding tests with the change so we can make sure it's working?

Yeah sure. I will work on getting those added today.

jeff-miller-cfa:
@Quanzzzz can you point me to an example of where tests are implemented? I see no unit tests in almost every package. I see a couple smoke tests but I'm not sure how you would verify that this is working with that method as you would have to verify that the multiple retries occurred.

dagnir:
Hi @jeff-miller-cfa you can refer to our V2 repo for examples of how tests are written: https://github.com/aws/aws-sdk-java-v2/tree/master/core/sdk-core/src/test/java/software/amazon/awssdk/core

the test cases under `retry` would probably be the best to look at. To make testing easier I'd recommend making the `StsRetryCondition` package private

jeff-miller-cfa:
> Hi @jeff-miller-cfa you can refer to our V2 repo for examples of how tests are written: https://github.com/aws/aws-sdk-java-v2/tree/master/core/sdk-core/src/test/java/software/amazon/awssdk/core
> 
> 
> 
> the test cases under `retry` would probably be the best to look at. To make testing easier I'd recommend making the `StsRetryCondition` package private

Ok thanks. Yeah I wasn't sure how you guys would feel about changing that scope but I can definitely do that. Unfortunately with the way the STS client is injected I would have to refactor a lot of code to get that class to be testable.

dagnir:
Thanks for adding the test cases! This looks good to me, I'll work on pulling in upstream.

jeff-miller-cfa:
> Thanks for adding the test cases! This looks good to me, I'll work on pulling in upstream.

Any update on getting this pulled in?

debora-ito:
@jeff-miller-cfa the change was merged upstream and was made available in today's release - `1.11.920`.
Thank you for your contribution.

PR #2465, opened 2020-12-04
Should be used Boolean.valueOf instead of Boolean.getBoolean

QiAnXinCodeSafe:
getBoolean  Returns true if and only if the system property named by the argument exists and is equal to the string "true".


debora-ito:
@QiAnXinCodeSafe The change was merged upstream and will be available in the next release. Thank you for the contribution!

PR #2467, opened 2020-12-04
The = = compare Integer should not be used

QiAnXinCodeSafe:
The = = compare Integer should not be used


QiAnXinCodeSafe:
compilation errors

PR #2473, opened 2020-12-14
Add hashcode and equals functions to S3 Bucket

smulikHakipod:
Adding `equals` and `hashcode` to S3 bucket

### DIFF
--- a/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/model/Bucket.java
+++ b/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/model/Bucket.java
@@ -171,4 +171,41 @@ public class Bucket implements Serializable {
         this.name = name;
     }
 
+
+    @Override
+    public boolean equals(Object obj) {
+        if (this == obj)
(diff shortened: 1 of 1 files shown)


smulikHakipod:
@debora-ito I see that you are active here, can you please have a look and see if I need to change anything? Sorry to bother you, Thanks!

debora-ito:
@smulikHakipod can you tell us more about the motivation for this change? In reality, S3 buckets names are unique across all AWS accounts.

Also, we try to avoid customizations in the low level classes.

smulikHakipod:
Sure, sorry for the delay. We want to store the buckets in a Set. We sometime brings the Bucket again, and its getting duplicated in the Set. All other resources we use dont have this kind problem, like EC2 instance, Lambda and so on.

debora-ito:
Hi there, closing this PR after a long silence, sorry about that. We don't have plans to merge this at this point.

I checked the SDK for Java v2 and confirmed it has equals and hashcode implemented, so please move to use v2 if possible.

PR #2609, opened 2021-07-14
Separate CHANGELOG into 1.11.x + current.

larrywest:
The CHANGELOG.md file has become so big that sometimes it is not rendered properly.

*(Not serious enough to merit an issue, this is simply a matter of readability and perhaps taste.)*

### Description of problem

In my Chrome and Firefox browsers, on MacOS, the CHANGELOG.md file had become so large (I assume this is the cause) this year that it often would only render in raw form, without applying the Markdown formatting.

Perhaps this is not wide-spread, but the start of 1.12.x is still an opportunity for a clean break.

### Description of changes

All changes from the 1.11.x range were moved to CHANGELOG-1.11.md.

Brief explanations were added to the top of that file, the bottom of CHANGELOG.md, and the "Release notes" section of the README. 

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice. (_Yes, I agree._)


debora-ito:
Hi @larrywest thank you for reaching out. Yes, we noticed the changelog stopped rendering the markdown format, the change you're proposing is very welcome.

I'll work to pull in the changes to our internal packages upstream, and will update here when I'm finished.

debora-ito:
@larrywest the change went out in yesterday's release. 1.11.x entries are now in a separate changelogs/CHANGELOG-1.11.x.md file, and the main CHANGELOG is rendering again. Thank you!

PR #2623, opened 2021-08-05
Update ReceiveMessageBatchTask messages to be stored and removed in fifo order

kmbotts:
*Issue #, if available:*
n/a

*Description of changes:*
Messages retrieved by spawned Receive task are removed in reverse order. Updated to have them polled in order as they were received. This was not intended to make the Amazon SQS Buffered Asynchronous Client fully support FIFO queues, but it's a step in that direction.


kmbotts:
Not necessarily in FIFO order, just the order they were received from queue. If that were to be a FIFO queue, then the order is preserved prior to fulfilling the future back to the caller.

debora-ito:
We don't have plans to merge this change. Closing this as a stale PR.

PR #2668, opened 2021-11-11
Improve MAVEN build Performance

hj987654321:
Maven allows you to run the compiler as a separate process by setting `<fork>true</fork>`. This feature can lead to much less garbage collection and make Maven build faster. This project has more than 1000 source files. We can consider enabling this feature.

Maven can recompile only the classes that were affected by a change. This feature is the default. We can activate it by setting `<useIncrementalCompilation>true</useIncrementalCompilation>`.

=====================
If there are any inappropriate modifications in this PR, please give me a reply and I will change them.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -385,6 +385,10 @@
             <target>${jre.version}</target>
             <encoding>UTF-8</encoding>
             <forceJavacCompilerUse>true</forceJavacCompilerUse>
+          	<fork>true</fork>
+          	<useIncrementalCompilation>true</useIncrementalCompilation>
+
+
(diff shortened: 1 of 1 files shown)


mpkorstanje:
Would recommend against merging. This user has spammed several projects with the same/similar PR. 

The PR is unlikely to make any improvement as noted here: https://github.com/apache/maven-surefire/pull/394#issuecomment-966250959

debora-ito:
Thank you for bringing this up @mpkorstanje.

We are not incorporating build changes to SDK 1.x at this time. Closing.

PR #2672, opened 2021-11-22
resovles #1648 set content type multipart upload

### LINKED ISSUE #1648: Multipart file upload content type
It looks like the normal `putObject` method on the S3 client will add a content-type header for you if one is missing. This is super helpful and I couldn't see this behaviour in the `initiateMultipartUpload` method. All the files I've uploaded end up having the content-type set to `application/octet-stream`

I'm currently using the AmazonS3 interface via the https://github.com/spring-cloud/spring-cloud-aws project and it's not possible to set the ObjectMetadata there.

Would it be possible to add this functionality to the `initiateMultipartUpload` method, as I can see the rest api supports that header?

JasonVoss:
Resolves #1648 

The issue is a feature request calling for automatically setting the content-type header of a multipart upload.  Similar to how the putObject method in the S3Client does it, by using the file extension.  Multipart uploads happen in 3 parts, initiate, send parts, and close.   The initiate does not take a file.  Initiate allows setting the content type in the object metadata.   I expect that initiateMultipartUpload will continue to determine content type from object metadata.   This seems like a fine way to do it, but I see there is still an open issue here, and is flagged for community help-wanted.   I added helper methods to set the content type given a file that will upload.  I kept the method name long and descriptive to show this is setting the content type in the object metadata, while there is still the existing setters to set the metadata.  Use of existing methods is not disrupted or changed in any way.

Also, I see there are no unit tests in the src code, but I also read in gitter that unit tests are welcome in pull requests.  I can remove those if needed.  

In the issue #1648, the question was because spring-cloud-aws was not allowing the user to set the content type.  This has since been fixed.  https://github.com/spring-cloud/spring-cloud-aws/issues/262.

### DIFF
--- a/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/model/InitiateMultipartUploadRequest.java
+++ b/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/model/InitiateMultipartUploadRequest.java
@@ -16,7 +16,9 @@ package com.amazonaws.services.s3.model;
 
 import com.amazonaws.AmazonWebServiceRequest;
 import com.amazonaws.services.s3.AmazonS3;
+import com.amazonaws.services.s3.internal.Mimetypes;
 
+import java.io.File;
 import java.io.Serializable;
 import java.util.Date;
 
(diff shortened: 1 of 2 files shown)


debora-ito:
Closing this as stale, please see my comment in https://github.com/aws/aws-sdk-java/issues/1648#issuecomment-2229605972.

Issue #2130, opened 2019-10-18
StateMachine.json() timestamp formatting error


jkoe-r7:
Step Function documentation (https://docs.aws.amazon.com/step-functions/latest/dg/amazon-states-language-wait-state.html) states `Timestamps must conform to the RFC3339 profile of ISO 8601, with the further restrictions that an uppercase T must separate the date and time portions, and an uppercase Z must denote that a numeric time zone offset is not present, for example, 2016-08-18T17:33:00Z.`, however when calling StateMachine.json() on a state machine which utilizes a timestamp there appears to be a formatting error because the joda time parser being used expects hh:mm:ss.fractionZ and is not parsing hh:mm:ssZ because it considers the 'Z' at that position to be invalid.

SDK version: 1.11.604

State machine to reproduce:
```
{
  "StartAt": "TestTime",
  "States": {
    "TestTime": {
      "Type": "Wait",
      "Timestamp": "2016-03-14T01:59:00Z",
      "Next": "HelloWorld"
    },
    "HelloWorld": {
      "Type": "Pass",
      "Result": "Hello World!",
      "End": true
    }
  }
}
```
Corresponding stack trace:
```
Caused by: com.fasterxml.jackson.databind.JsonMappingException: Invalid format: "2016-03-14T01:59:00Z" is malformed at "Z" (through reference chain: com.amazonaws.services.stepfunctions.builder.StateMachine$Builder["States"]->java.util.LinkedHashMap["TestTime"]->com.amazonaws.services.stepfunctions.builder.states.WaitState$Builder["Timestamp"])
	at com.fasterxml.jackson.databind.JsonMappingException.wrapWithPath(JsonMappingException.java:394)
	at com.fasterxml.jackson.databind.JsonMappingException.wrapWithPath(JsonMappingException.java:353)
	at com.fasterxml.jackson.databind.deser.BeanDeserializerBase.wrapAndThrow(BeanDeserializerBase.java:1711)
	at com.fasterxml.jackson.databind.deser.BeanDeserializer.deserializeWithUnwrapped(BeanDeserializer.java:643)
	at com.fasterxml.jackson.databind.deser.BeanDeserializer.deserializeFromObject(BeanDeserializer.java:321)
	at com.fasterxml.jackson.databind.deser.BeanDeserializer._deserializeOther(BeanDeserializer.java:194)
	at com.fasterxml.jackson.databind.deser.BeanDeserializer.deserialize(BeanDeserializer.java:161)
[...]

PR #2683, opened 2021-12-10
Fix issue #2130

Allenzhanghx:
[*Issue #2130 ]

*Description of changes:*
Change StateMachine.Builder fromJson() so that timestamp with format HH:MM:SSZ will be changed to timestamps with format HH:MM:SS.fractionZ.


debora-ito:
@Allenzhanghx we appreciate the PR, but won't be able to merge it.

We don't have plans to change this feature right now in Java SDK v1, and besides, changing the timestamp to add the fractional zeros is a potential breaking change. I'll go ahead and close this.

PR #2725, opened 2022-03-22
fix: specify how to configure Default location

froblesmartin:
There is no place specifying how to configure a geolocation record with Default location. I have checked other SDKs and it is not mentioned anywhere.


debora-ito:
Hi @froblesmartin thank you for submitting the PR. We won't be able to merge it unfortunately.

GeoLocation is an auto-generated class (note the `@Generated` annotation on the top of the class name) based on a json model. Any change must be made in the model otherwise the next update will overwrite it. The model is maintained upstream by the service team so I'll contact them internally with your suggestion.

Thank you!

Issue #2733, opened 2022-04-10
DefaultTimeoutThreadPool builded with hard-coded params


kurenchuksergey:
### Describe the feature

Unfortunately, nowadays I can't customise default executor and it strange
Executor should be customised

### Use Case

For example I can use pooling model with scheduled  scan. In that case if I will scan it rare then 5 second (like once per 10 second)  in each time I  will produce new thread. But it's not needed and also distracts from real problems when you make jfr/monitorings

### Proposed Solution

just give possibility to setup my own values

### Acknowledgements

- [X] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### AWS Java SDK version used

2

### JDK version used

any

### Operating System and version

any

debora-ito:
Hi @kurenchuksergey thank you for spending the time to submit the PR. We won't be able to accept it unfortunately.

`TimeoutThreadPoolBuilder` is meant to be used as an internal tool (`@SdkInternalApi`), having a configurable timeout is not in the plans for it.

Will go ahead and close this. Feel free to reach out if you have other questions about the Java SDK.

PR #2734, opened 2022-04-10
 #2733 - DefaultTimeoutThreadPool builded with hard-coded params

kurenchuksergey:
Describe the feature
Unfortunately, nowadays I can't customise default executor and it strange
Executor should be customised

Use Case
For example I can use pooling model with scheduled scan. In that case if I will scan it rare then 5 second (like once per 10 second) in each time I will produce new thread. But it's not needed and also distracts from real problems when you make jfr/monitorings

*Issue #, if available:*

*Description of changes:*


kurenchuksergey:
fixes #2733

debora-ito:
See my comment in #2733

PR #2853, opened 2022-09-23
Match no_proxy environment variable behaviour with other tools

hpryce:
*Description of changes:*

In curl, wget, Ruby, Python, Go and others the no_proxy variable matches subdomains as well as the parent, as described in this article:
https://about.gitlab.com/blog/2021/01/27/we-need-to-talk-no-proxy/

The Java nonProxyHosts setting works differently, needing explicit wildcard characters added as per:
https://docs.oracle.com/javase/8/docs/technotes/guides/net/proxies.html

Especially with path-style URLs for S3 being deprecated it is helpful to be able to match the behaviour of other tools and libraries on hosts when trying to set all S3 buckets in a region to bypass any proxy set.

While this change may impact behaviour, as requests which previously hit a proxy might now be excluded, the original addition of respecting the no_proxy environment variable came in 2019 and was likewise a change in behaviour - as such I believe it is acceptable to make this change even given the potential impact on clients.

### DIFF
--- a/aws-java-sdk-core/src/main/java/com/amazonaws/ClientConfiguration.java
+++ b/aws-java-sdk-core/src/main/java/com/amazonaws/ClientConfiguration.java
@@ -1109,7 +1109,16 @@ public class ClientConfiguration {
     private String getNonProxyHostsEnvironment() {
         String nonProxyHosts = getEnvironmentVariableCaseInsensitive("NO_PROXY");
         if (nonProxyHosts != null) {
-            nonProxyHosts = nonProxyHosts.replace(",", "|");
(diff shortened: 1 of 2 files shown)


zoewangg:
Hi @hpryce thank you for your interest in contributing to our project. We have discussed internally and unfortunately, we will not be able to accept this change because it will likely break customers, and although this behavior matches with other tools, it is inconsistent with the Java behavior, which could cause confusion. The environment variable is a different story; it was an addition, so there was no breaking change introduced. On the hindsight, we should've probably made environment variable behave the same way as the system property.

We will close the PR. We appreciate your time and effort spent on this!

Issue #2889, opened 2022-12-14
`doesBucketExistV2` breaks interoperability with other S3 compatible APIS


pwharned:
### Describe the bug

The AWS S3 APIS are more or less interoperable with other S3 compatible apis. However the implementation of some methods breaks the interoperability of client apps that use the SDK to interoperate with S3 compatible apis( for example Hive S3 integrations). (and are besides smelly)  The `doesBucketExistV2` method seems to make a call to the AWS ACL apis? Which can be seen in the implementation.

```

    @Override
    public boolean doesBucketExistV2(String bucketName) throws SdkClientException {
        try {
            ValidationUtils.assertStringNotEmpty(bucketName, "bucketName");
            getBucketAcl(bucketName);
            return true;
        } catch (AmazonServiceException ase) {
            // A redirect error or an AccessDenied exception means the bucket exists but it's not in this region
            // or we don't have permissions to it.
            if ((ase.getStatusCode() == Constants.BUCKET_REDIRECT_STATUS_CODE) || "AccessDenied".equals(ase.getErrorCode())) {
                return true;
            }
            if (ase.getStatusCode() == Constants.NO_SUCH_BUCKET_STATUS_CODE) {
                return false;
            }
            throw ase;
        }
    }
```

In this trivial example replacing the AWS endpoint with a GCS endpoint I can easily list buckets in my GCS account.

```
import com.amazonaws.auth.AWSStaticCredentialsProvider
import com.amazonaws.client.builder.AwsClientBuilder
import com.amazonaws.regions.Regions
import com.amazonaws.services.s3.model.ObjectListing
import com.amazonaws.services.s3.{AmazonS3, AmazonS3Client, AmazonS3ClientBuilder, AmazonS3URI}
import com.amazonaws.services.s3.model.ObjectListing

object Main extends App{

  import com.amazonaws.auth.AWSCredentials
  import com.amazonaws.auth.BasicAWSCredentials

  val endpointConfig = new AwsClientBuilder.EndpointConfiguration("https://storage.googleapis.com", "us-east1")

  val credentials = new BasicAWSCredentials("", "")
  val s3client: AmazonS3ClientBuilder  = AmazonS3ClientBuilder
    .standard()
  s3client.setEndpointConfiguration(endpointConfig);
  val finalClient = s3client.withCredentials(new AWSStaticCredentialsProvider(credentials))
    .build();

  import com.amazonaws.services.s3.model.Bucket

[...]

PR #2890, opened 2022-12-14
fix doesBucketsExistV2

pwharned:
current implementation doesn't make sense and also breaks interoperability with other S3 compatible apis

*Issue #2889 *

*Description of changes:*

Instead of calling ACL apis, we should just list the buckets and check whether the user provided bucket name matches one of the buckets in the list. See #2889  for full details regarding the justification for this change. 


pwharned:
closes #2889

debora-ito:
See my comment in #2889.

PR #2925, opened 2023-02-19
Allow to customize QueueBufferConfig instances

flozano:
I would like to use a version of QueueBufferConfig that can be reconfigured. QueueBufferConfig is not final, so with just this small change I can hook what I need to allow things like `maxBatchOpenMs` to be dynamically redefined for existing/new buffers.

*Description of changes:*

Just allow to override QueueBufferConfig copy instantiation.


debora-ito:
We don't have plans to merge this change. Closing this as a stale PR.

flozano:
what's the harm in doing so? as an AWS customer who spends quite a lot of money in SQS, how can I achieve what I need (dynamically handle maxBatchOpenMs) ?

PR #2930, opened 2023-02-22
remove whitespace from cloudfront signed URL policy

romed:
I removed a whitespace character from the string used to create a policy for Cloudfront signed URLs.
According to the Docs and the Troubleshooting guide: https://aws.amazon.com/premiumsupport/knowledge-center/cloudfront-troubleshoot-signed-url-cookies/ there should not be whitespace characters in there as the validation will fail.


debora-ito:
I think that specific whitespace is not a problem, the policy statement examples in the CloudFront Dev Guide have a space between "Statement:" and "[": 
https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/private-content-creating-signed-url-custom-policy.html#private-content-custom-policy-statement-examples

Closing this as stale.

PR #2961, opened 2023-04-15
update: change the default thread pool for AwsAsyncClientBuilder.

ChoiJunsik:
*Issue #, if available:*

In the case of fixedThreadPool, unnecessary thread occupancy and memory leak can occur. That's why I'm going to change the default thread pool implementation of AwsAsyncClientBuilder.

In my case, the thread did not end after the process of transmitting large amounts of data to SQS, resulting in a memory leak.

image

image

*Description of changes:*

```java
private ExecutorService defaultExecutor() {
            return new ThreadPoolExecutor(0, getClientConfiguration().getMaxConnections(), 60L, TimeUnit.SECONDS, new LinkedBlockingQueue<Runnable>());
        }
```


debora-ito:
@ChoiJunsik we appreciate your PR, but we don't intend to change the behavior of the default async executor.

In your code you can provide a custom executor in the SQS client builder, as shown in our Developer Guide - https://docs.aws.amazon.com/sdk-for-java/v1/developer-guide/creating-clients.html#creating-async-clients.

PR #2969, opened 2023-04-27
Remove ion-java

jobarr-amzn:
*Description of changes:* Java SDK v2 dropped support for Ion, and it’s unsupported for v1. The version of ion-java (`software.amazon.ion:ion-java:1.0.2`) bundled with aws-sdk-java is substantially behind the current version (`com.amazon.ion:ion-java:1.9.6`) and uses a different java package. This causes customers using aws-sdk-java to get an out-dated version in their dependencies. It would be better to have an up-to-date version or no transitive dependency so that customers don't miss out on the bug fixes and  performance improvements ion-java has made since then.

This commit removes ion-java from aws-sdk-java, removing also all the code that depends on it.
If it is preferable, I also have a patch ready which upgrades the ion-java version instead.


zdenda-online:
Thanks, I think removal is a better option if it is not supported.

Any ETA to merge and release a new version? (whoever is in charge :grin: )

jumraiya:
+1 on getting this merged.

davidh44:
Ion has been removed from Java v1

PR #3012, opened 2023-07-26
Removing the jackson backward incompatible dependency on class (only present on jackson 2.12 onwards)

pranavbhole:
*Issue #, if available:* AWS-java-sdk-core is incompatible with jackson 2.10.5 and older 

*Description of changes:*

Hello,
Creating the PR to make Aws-java-sdk core backward compatible with the older jackson version. Unfortunately com.fasterxml.jackson.databind.PropertyNamingStrategies is not present on jackson version 2.10.5 backwards and it makes the aws lib incompatible with older version of jackson.
Tested the change on local env. Other Unit tests are already covering the change.

Can you please review ?

Thank you

### DIFF
--- a/aws-java-sdk-core/src/main/java/com/amazonaws/util/EC2MetadataUtils.java
+++ b/aws-java-sdk-core/src/main/java/com/amazonaws/util/EC2MetadataUtils.java
@@ -29,7 +29,6 @@ import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.databind.DeserializationFeature;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
-import com.fasterxml.jackson.databind.PropertyNamingStrategies;
 import com.fasterxml.jackson.databind.PropertyNamingStrategy;
 import java.net.URI;
 import java.util.Arrays;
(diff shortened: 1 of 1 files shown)


debora-ito:
This is covered by the catch -

https://github.com/aws/aws-sdk-java/blob/8cade0bc9cf72071bae24fd4d85c3ae182b7ed0f/aws-java-sdk-core/src/main/java/com/amazonaws/util/EC2MetadataUtils.java#L87-L93

I don't think we can make this change, since the SDK is using jackson 2.12.6.

pranavbhole:
@debora-ito AWS-SDK correctly does not work with the containers which has older version of the jackson in the class path. 
This change will make the sdk compatible with older version. 
For example: Hadoop mapreduce job is failing to instantiate the latest AWS sdk as container is trying to load classes from newer jackson which are not present.

this change is simple, just removing the newer references of the class.
Please let me know if you need more information.

debora-ito:
@pranavbhole can you open a new github issue and give as much detail as possible?
I want to better understand the issue, maybe we can fix it in another way.

pranavbhole:
@debora-ito I have created sample app to simulate what my hadoop map reduce container is doing. https://github.com/pranavbhole/aws-sdk-usage-example/tree/main  
base app has older jackson than what sdk is pulling.
you can run , 
`mvn clean install exec:java -Dexec.mainClass="org.pra.aws_example.App"`

debora-ito:
@pranavbhole I appreciate the code sample, but in the `new issue template` we ask you to provide other artifacts, like logs and stacktraces. Those are helpful for us to understand how the error manifests in your environment.

debora-ito:
Closing this PR. If you want us to investigate the original problem, please open a Github issue.

PR #3185, opened 2025-07-14
Update VersionInfoUtils - Fix deprecation message logic to use OR instead of AND

maxpdr96:
Fix deprecation message logic to use OR instead of AND

### Motivation
The deprecation warning message was previously suppressed only if both the environment variable and the system property were set to 'true', due to the use of `&&`.

This PR corrects the condition to use `||`, so the warning is suppressed if **either** the environment variable or the system property is explicitly set to 'true'.

### Changes
- Replaced `&&` with `||` in the deprecation announcement check.


maxpdr96:
Hi @debora-ito ,
Thank you for the feedback. It's just that the warning message:

"This message can be disabled by setting the " + DISABLE_DEPRECATION_ANNOUNCEMENT_ENV_VAR
" environment variable or " + DISABLE_DEPRECATION_ANNOUNCEMENT_SYS_PROP + " system property to 'true'."

makes it sound like setting either one would be enough because the "or" . However, the current code requires both to be set to true, which is why I suggested the change. But I appreciate your clarification and the response.

